PR #44, opened 2024-08-16
record_ui: add link to inspire for datasets that have publications

psaiz:
The landing page of a record will make a call to inspire to see if there are publications for this DOI. If so, it will add a link to it.


psaiz:
If there are no publications:
Screenshot 2024-08-16 at 11 23 46

In case the call to inspire returned some entries (the DOI had been overwritten by hand for this screenshot)
Screenshot 2024-08-16 at 11 24 11

zzacharo (approved):

PR #55, opened 2024-09-09
Documentation fixes

Biscgit:
Fixed docker compose commands and added name to web container

### DIFF
--- a/scripts/start-server-debug.sh
+++ b/scripts/start-server-debug.sh
@@ -1,14 +1,18 @@
 #!/bin/bash
+
 echo "Ready to start the web server in debug mode"
-cd /code
-export INVENIO_SECRET_KEY=$(head -c 500 /dev/urandom | tr -dc 'a-zA-Z0-9~!@#$%^&*_-' | fold -w 10 | head -n 1)
+cd /code || exit 1
+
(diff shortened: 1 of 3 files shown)


scripts/start-server-debug.sh
```diff
@@ -1,14 +1,18 @@
...
+export INVENIO_SECRET_KEY
+
 if [ -d "/opt/invenio/var/instance/python/lib/python3.9/site-packages/cernopendata" ]; then
   echo "The installation directory is still there... let's overwrite it"
   cp /code/cernopendata/setup.py /code/setup.py
   pip install -e .
 
   if [ -d "/cold" ]; then
-    cd /cold
+    cd /cold || exit 1
```
zzacharo, line 15: Out of curiosity: why the need for `exit 1`? If I understand correctly, the `if` statement above checks for the existence of the folder.

zzacharo (commented):
Well done for your first PR 🚀

scripts/start-server-debug.sh
Biscgit: Thanks for your response!
I know that it checks for the folder's existence first (and permission errors should be almost nonexistent), but as I run `shellcheck` over it, it kept complaining there. I thought there won't be any harm by adding it. But you can also remove it, that's fine!

zzacharo (approved):

PR #61, opened 2024-09-26
New script

Biscgit:
Created a script to easily get the filename by a record's ID, so it can be easily found and inserted with `cernopendata`.

### DIFF
--- a/scripts/recordid-to-filepath.sh
+++ b/scripts/recordid-to-filepath.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-# This script returns the exact path of a record
-# Pass the id as the first argument
-# On finding the correct file, it prints its path to the console
+# This script returns the exact path of a record.
+# Pass the id as the first argument.
+# On finding the correct file, it prints its path to the console.
 
 if [ $# -eq 0 ]; then
   echo "Error: No record ID provided."
--- a/scripts/recordid-to-filepath.sh
+++ b/scripts/recordid-to-filepath.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# This script returns the exact path of a record
+# Pass the id as the first argument
+# On finding the correct file, it prints its path to the console
+
+if [ $# -eq 0 ]; then
+  echo "Error: No record ID provided."
+  exit 1
+fi
+
+result=$(
+  grep -ri "$1" /content/data/records/ |
+  grep  '"recid"' |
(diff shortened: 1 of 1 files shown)


scripts/recordid-to-filepath.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
tiborsimko, line 1: Is such a script necessary? 

I'm using a simple one-liner to achieve the same needs:

```console
$ rg '^    "recid": "123"' data/records
• data/records/cms-hlt-configuration-files-2010.json
732:    "recid": "123",
```

Note also the leading space formatting in the regexp, which helps to avoid false positives, because record ID can be referenced in links. 

(This technique works with less false positives, because all the fixture files are formatted in the same way, using `scripts/clean_json_file.py`.)

Biscgit: As far as I've seen, `Ripgrep` is not installed on the shipped web container by default (where following the opendata development guide, all commands for setup and working with records are executed).

With the records I have tested, my script should also avoid false positives (order by leading whitespaces). 

I think the command is quite important for developing and fixing issues with certain records, so if not as a separate script, this or yours should be at least added to the dev documentation somewhere.

scripts/recordid-to-filepath.sh
```diff
@@ -0,0 +1,20 @@
...
+# On finding the correct file, it prints its path to the console.
+
+if [ $# -eq 0 ]; then
+  echo "Error: No record ID provided."
+  exit 1
+fi
+
+result=$(
+  grep -ri "$1" /content/data/records/ |
+  grep  '"recid"' |
```
psaiz, line 14: this grep would catch recid references inside the document, like `pileup->links->recid`. It might be better to look for regular expresion (like `grep -e '^    "recid":'`

Biscgit: I think then it is better to just include the command mentioned on the top in the dev documentation. :+1:

Mine worked with multiple random scripts I tested, that's why I thought it should work fine, but it does not on some others now I've seen.

scripts/recordid-to-filepath.sh
tiborsimko: If `rg` is not available, you could also use `git grep` or `grep -r`.

(Personally I'm always grepping locally, i.e. outside of the container.)

PR #62, opened 2024-09-26
file_indices: Fix issue getting the indices from a file

psaiz:
record_view: Fixed issue on how to access the file_indices of a record.

### DIFF
--- a/cernopendata/modules/records/utils.py
+++ b/cernopendata/modules/records/utils.py
@@ -48,8 +48,8 @@ import re
 def get_file_index(pid, record, file_index, **kwargs):
     """Return the list of entries."""
     entry_name = file_index.replace(".txt", ".json")
-    file_indices = record.get("_file_indices", [])
-    for entry in file_indices:
+
+    for entry in record.file_indices:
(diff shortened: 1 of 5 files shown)


cernopendata/modules/records/utils.py
```diff
@@ -162,8 +162,8 @@ def record_file_page(pid, record, page=1, **kwargs):
     except Exception:
         items_per_page = 5
 
-    _files = record.files
-    index_files = record._file_indices
+    _files = record.get("files", [])
```
zzacharo, line 165: I think the problem is https://github.com/cernopendata/cernopendata-portal/blob/853b55af0554605913fa1005afc56142a9ed2ba6/cernopendata/config.py#L205 we should point to the new API class i.e. [RecordFilesWithIndex](https://github.com/cernopendata/cernopendata-portal/blob/853b55af0554605913fa1005afc56142a9ed2ba6/cernopendata/api.py#L47)

psaiz: Thanks!  Let me check that

zzacharo (approved):

PR #82, opened 2024-11-14
Option to exclude experiments from view

Biscgit:


Biscgit:
Tests should now also be done :+1:

psaiz (approved):

PR #93, opened 2024-11-27
File box

psaiz:
Closes https://github.com/cernopendata/opendata.cern.ch/issues/3600

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/records/FilesBoxApp.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/records/FilesBoxApp.js
@@ -69,7 +69,7 @@ const FilesBoxApp = () => {
     <>
       {items.total > 0 && (
         <>
-          <h3>{title}</h3>
+          <h2>{title}</h2>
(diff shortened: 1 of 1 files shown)


Biscgit (approved):
Looking good :+1:

cernopendata/modules/theme/assets/semantic-ui/js/records/FilesBoxApp.js
```diff
@@ -69,7 +69,7 @@ const FilesBoxApp = () => {
     <>
       {items.total > 0 && (
         <>
-          <h3>{title}</h3>
+          <h2>{title}</h2>
```
zzacharo, line 72: Is this change just intended to increase the font size? Maybe we can evaluate some CSS to do that? I am not sure from a11y point of view if `h2` is recommended in this case...

psaiz: The goal was to make it consistent with the other sections of that page, which are h2:
BEFORE:
Screenshot 2024-11-29 at 10 09 10
AFTER:
Screenshot 2024-11-29 at 10 09 54

PR #102, opened 2024-11-29
Security fixes

Biscgit:
Fixed a critical and a few high vulnerabilities affecting build and a test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,14 +108,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      # - name: Setup node
-      #   uses: actions/setup-node@v1
-      #   with:
-      #     node-version: "14"
-
-      # - name: Install Node dependencies
-      #   run: npm install -g jsonlint
-
       - name: Check fixtures
         run: ./run-tests.sh --check-fixtures
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,13 +108,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup node
-        uses: actions/setup-node@v1
-        with:
-          node-version: "14"
+      # - name: Setup node
+      #   uses: actions/setup-node@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -108,13 +108,13 @@ jobs:
...
-          node-version: "14"
+      # - name: Setup node
+      #   uses: actions/setup-node@v1
+      #   with:
+      #     node-version: "14"
 
-      - name: Install Node dependencies
-        run: npm install -g jsonlint
+      # - name: Install Node dependencies
+      #   run: npm install -g jsonlint
```
psaiz, line 117: Would. you mind removing the lines instead of commenting them out?

Dockerfile
```diff
@@ -45,8 +46,8 @@ RUN echo "Will install xrootd version: $XROOTD_VERSION (latest if empty)" && \
     yum install -y xrootd-"$XROOTD_VERSION" python3-xrootd-"$XROOTD_VERSION" && \
     yum clean -y all
 
-RUN pip uninstall pipenv -y && pip install --no-cache-dir --upgrade pip==20.2.4 setuptools==68.2.2 wheel==0.36.2 && \
-    npm install -g --unsafe-perm node-sass@6.0.1 clean-css@3.4.24 requirejs@2.3.6 uglify-js@3.12.1 jsonlint@1.6.3 d3@6.3.1 \
+RUN pip uninstall pipenv -y && pip install --no-cache-dir --upgrade pip==24.3.1 setuptools==70.0.0 wheel==0.45.1 && \
```
psaiz, line 49: I like that we are going for newer versions. Are those the latest available? If so, would it make sense to change it to require `=>` instead of `==`?

Biscgit: `pip` and `wheel` are the latest right now, I cannot set the version of `setuptools` to `71.0.0 or higher` because of other issues during the build (latest there is `75.6.0`)

Biscgit: Also, I just fixed that, but the dockerfile-linter does not appreciate the command like this

psaiz (approved):

PR #103, opened 2024-12-03
Portal anniversary

Biscgit:
Added small banner and animations for celebrating

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/anniversary.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/anniversary.js
@@ -53,7 +53,7 @@ function Celebrate() {
     )
 }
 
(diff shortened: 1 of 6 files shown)


cernopendata/modules/theme/assets/semantic-ui/js/anniversary.js
```diff
@@ -0,0 +1,59 @@
+import React, {useState} from "react";
+import ReactDOM from "react-dom";
+
+import Confetti from 'react-confetti'
+
+
+const unlock_date = new Date("2024-12-09")
```
tiborsimko, line 7: Better to specify Tuesday 10th 17:00 CET for unlocking, so that we have time to iterate over the blog post on Monday...

PR #109, opened 2024-12-17
Add RSS endpoint for news

Biscgit:
For implementing a news banner in Grafana on the landing page, it would be a nice addition to display the news from the main Open Data page. The returned XML should conform the RSS standards, similar to Grafana's default feed [xml file](https://grafana.com/blog/news.xml).


Biscgit:
This way we can display the latest news directly in the dashboard for curators and more. 
Others from outside could also use this to gather news from CERN Open Data

image

psaiz (approved):

PR #125, opened 2025-01-09
View: Return the experiment of a record page as a custom HTTP header alongside the rendered template

Biscgit:


Biscgit:
Using this, we can more easily enrich the logs with the corresponding experiments to records.
image

cernopendata/modules/records/utils.py
```diff
@@ -186,6 +187,19 @@ def record_file_page(pid, record, page=1, **kwargs):
     return jsonify({"total": rf_len, "files": paged_files})
 
 
+def _extract_experiment_name(record):
+    try:
+        experiment = str(record["collaboration"]["name"])
```
psaiz, line 192: Why `collaboration.name` instead of `experiment`?

cernopendata/modules/records/utils.py
```diff
@@ -186,6 +187,19 @@ def record_file_page(pid, record, page=1, **kwargs):
...
+        return experiment.split(" ", maxsplit=1)[0]
```
psaiz, line 193: If the entry belongs to multiple experiments, then only the first one will be in the headers. That doesn't seem right.

cernopendata/modules/records/utils.py
```diff
@@ -186,6 +187,19 @@ def record_file_page(pid, record, page=1, **kwargs):
...
+
+    except (KeyError, IndexError):
+        return "other"
+
+    except Exception as e:
+        logging.error(f"[{type(e)}] Could not extract experiment from record: {e}")
+        return "other"
```
psaiz, line 200: I would return a different value in each exception.

cernopendata/modules/records/utils.py
```diff
@@ -203,16 +217,22 @@ def record_metadata_view(pid, record, template=None):
...
+        )
     )
 
+    # include experiment in the header for direct annotation in the logs
+    response.headers["X-Record-Experiment"] = _extract_experiment_name(record)
+    return response
+
 
 def term_metadata_view(pid, record, template=None):
     """Term detail view."""
```
psaiz, line 238: What about enriching also the `doc_metadata_view` with the experiment?

cernopendata/modules/records/utils.py
Biscgit: The issue that I saw is, that by using the “experiments” variable, we get an array of possible different experiments, which makes the later analysis much more difficult (regarding that only about 5 records have more than one value set), so it was easier to extract the value from the collaborations name. 
Now that I see it, it might be better to implement it similarly to the SQL Queries we are using for MONIT SPQ. In those, we set multiple experiments to one keyword (i.e., `multiple`).

cernopendata/modules/records/utils.py
Biscgit: In this case, no, because that record variable holds a string with one exact value. The split is to extract the experiment's name, because (for CERN experiments) it is saved in the form of `[Experiment] Collaboration` (i.e., not including the *collaboration* in the string).

PR #129, opened 2025-01-13
stats: collect stats on record views and file downloads.

psaiz:


cernopendata/config.py
```diff
@@ -133,6 +137,66 @@
...
+STATS_QUERIES = {
+    "record-view": {
+        "cls": TermsQuery,
+        "permission_factory": None,
+        "params": {
+            "index": "stats-record-view",
+            "doc_type": "record-view-day-aggregation",
+            "copy_fields": {
+                "recid": "recid",
+                "parent_recid": "parent_recid",
```
zzacharo, line 153: I don't think you have a `parent_recid` no?

cernopendata/config.py
```diff
@@ -133,6 +137,66 @@
 # Celery
 CELERY_ACCEPT_CONTENT = ["json", "msgpack", "yaml"]
 
+STATS_EVENTS = EVENTS_CONFIG
+
+STATS_AGGREGATIONS = AGGREGATIONS_CONFIG
+
+STATS_QUERIES = {
```
zzacharo, line 144: I think you miss the `params.index_interval`. Does it make sense to already move it to yearly?

cernopendata/config.py
psaiz: Thanks, removed

cernopendata/config.py
psaiz: Thanks, I'll put it

zzacharo (approved):

Issue #128, opened 2025-01-13
fixtures: prettify output message structure


tiborsimko:
### Current behaviour

When uploading records, the current output of the `cernopendata fixtures` command does not look very well structured:

```text
record 69734 updated
2025-01-12 16:59:04.555600 This is an index file. Let's check the entries that it has: root://eospublic.cern.ch//eos/opendata/cms/mc/RunIISummer20UL16NanoAODv9/WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1/file-indexes/CMS_mc_RunIISummer20UL16NanoAODv9_WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8_NANOAODSIM_106X_mcRun2_asymptotic_v17-v1_120000_file_index.json
The file index contains 4 entries.
File index created
2025-01-12 16:59:04.682527 This is an index file. Let's check the entries that it has: root://eospublic.cern.ch//eos/opendata/cms/mc/RunIISummer20UL16NanoAODv9/WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1/file-indexes/CMS_mc_RunIISummer20UL16NanoAODv9_WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8_NANOAODSIM_106X_mcRun2_asymptotic_v17-v1_130000_file_index.json
The file index contains 1 entries.
File index created
2025-01-12 16:59:04.733111 This is an index file. Let's check the entries that it has: root://eospublic.cern.ch//eos/opendata/cms/mc/RunIISummer20UL16NanoAODv9/WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1/file-indexes/CMS_mc_RunIISummer20UL16NanoAODv9_WJetsToLNu_HT-800To1200_TuneCP5_13TeV-madgraphMLM-pythia8_NANOAODSIM_106X_mcRun2_asymptotic_v17-v1_270000_file_index.json
The file index contains 19 entries.
File index created
[...]

Issue #117, opened 2025-01-06
Add a new field: availability


psaiz:
Add a new optional field to the records that would identify the status of the dataset. Possible status are:
* Available
* Sample available
* Needs to be requested

tiborsimko:
We already have a `distribution.availability` field that is meant to be used precisely for this purpose?

```json
    "distribution": {
      "properties": {
        "availability": {
          "description": "Specifies if dataset is on-demand, online in future might be also tape, disk",
          "enum": [
            "ondemand",
            "online"
          ],
          "type": "string"
        },
```

psaiz:
Even better, thanks! Then, to complete this ticket it looks like the only change will be to add new entries to the `enum`

psaiz:
Done and deployed

PR #133, opened 2025-01-17
(#128, #117)  Improve logging and add availability field 

Yuvrajsinghspd09:
1. Improved logging output structure for better readability and maintainability (#128).
2. Added an optional "availability" field to dataset records with statuses: `Available`, `Sample available`, and `Needs to be requested` (#117).


psaiz:
Hi @Yuvrajsinghspd09. 
Thanks for the pull request. One of the two tasks that you picked (#117) is a bit trickier, and it needs more internal discussions. If you don't mind, I would prefer to leave that one out of the scope of this PR. 
The PR also seems to add some files that are not necessary. Could you please check that?
Finally, would you mind checking also the comments about the verbose level (changing the logging to DEBUG), and the final summary?

Thanks again,
pablo

PR #137, opened 2025-01-30
cold_data: Store multiple uri for a single file.

psaiz:
Adding a new class inheriting from FileObject that allows two urls for an entry. It also calculates the availability of a file (depending if the `hot` url has a value`), and for a record, which goes through all the files in that record. 

The availability for a record is defined with the following formula:
```
If the record has no files, and it has the field 'distribution.availability':
    avl = record.distribution.availability #At the moment, this is used for ondemand
if the record has files:
   if all the files are in the same state:
       avl = record.file[0].availability. # If all the files are online, or all ondemand
   else
      avl = 'sample files'
```

### DIFF
--- a/cernopendata/modules/fixtures/cli.py
+++ b/cernopendata/modules/fixtures/cli.py
@@ -44,10 +44,11 @@ from invenio_pidstore.models import PersistentIdentifier
 from invenio_records import Record
 from sqlalchemy.orm.attributes import flag_modified
 
-from cernopendata.api import FileIndexMetadata, RecordFilesWithIndex
+from cernopendata.api import FileIndexMetadata, MultiURIFileObject, RecordFilesWithIndex
 from cernopendata.modules.records.minters.docid import cernopendata_docid_minter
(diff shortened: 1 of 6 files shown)


cernopendata/modules/fixtures/cli.py
```diff
@@ -80,27 +85,27 @@ def _handle_record_files(record, data):
...
-            print("File index created")
             f.delete()
         elif "type" in file and file["type"] == "index.txt":
             # The txt indexes should be ignored. Just delete the file
             f.delete()
         else:
             real_files.append(file)
             try:
-                obj = ObjectVersion.create(record.bucket, filename, _file_id=f.id)
+                obj = FileObjectCold.create(record.bucket, filename, f.id)
```
psaiz, line 102: From https://github.com/cernopendata/cernopendata-portal/pull/138
```suggestion
                obj = FileObjectCold.create_version(record.bucket, filename, f.id)
```

zzacharo (approved):

Issue #118, opened 2025-01-06
Change the availability facet


psaiz:
Currently there is a facet for the `on-demand` entries.

![Image](https://github.com/user-attachments/assets/1d06fe63-09bd-49ed-b857-ba85f3763ac2)

Those records have a message like:

![Image](https://github.com/user-attachments/assets/3e9501b3-32d2-47f1-b645-efb93311717f)

See https://opendata-qa.cern.ch/record/13066 as an example

The idea behind is very similar to what should happen with the entries that are in cold-storage, where the users have to request that some datasets should be made available. 

As a first step, let's change the facet. Instead of a toggle, let's put a menu that looks for the `availability` field 
in the dataset (see #117)

psaiz:
Removing the `availability` box with a toggle, and adding one with multiple options 
![Image](https://github.com/user-attachments/assets/63d86d35-f822-440c-bbc2-ded3963d7113)
![Image](https://github.com/user-attachments/assets/adb638d3-d6cc-4c49-97e3-084742e222d1)

psaiz:
![Image](https://github.com/user-attachments/assets/265206ad-a386-4d4e-bdae-115f28e77ab2)

PR #138, opened 2025-02-03
Availability facet

psaiz:
Addresses https://github.com/cernopendata/cernopendata-portal/issues/118

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -356,6 +356,8 @@ def _query_parser_and(qstr=None):
         )
     else:
         _query = dsl.Q()
+    # This is only for backward compatibility, in case a link like https://opendata.cern/search?q=&f=ondemand
+    # had been saved. Since it is kind of a corner case, we could consider dropping it completely on a future release
     if (
         request.values.get("ondemand") != "true"
         and request.values.get("ondemand") != "ondemand"
@@ -479,6 +481,7 @@ RECORDS_REST_FACETS = {
                 ),
             ),
             experiment=dict(terms=dict(field="experiment", order=dict(_key="asc"))),
+            availability=dict(terms=dict(field="availability")),
             year=dict(
                 terms=dict(field="date_created", size=70, order=dict(_key="asc"))
             ),
@@ -527,6 +530,7 @@ RECORDS_REST_FACETS = {
             keywords=dict(terms=dict(field="keywords", order=dict(_key="asc"))),
         ),
         "post_filters": dict(
+            availability=terms_filter("availability"),
             type=nested_filter("type.primary", "type.secondary"),
             experiment=terms_filter("experiment"),
             year=terms_filter("date_created"),
--- a/cernopendata/modules/fixtures/data/terms/terms.json
+++ b/cernopendata/modules/fixtures/data/terms/terms.json
@@ -35,6 +35,17 @@
     }
   },
   {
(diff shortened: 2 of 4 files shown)


cernopendata/config.py
```diff
@@ -355,11 +355,12 @@ def _query_parser_and(qstr=None):
         )
     else:
         _query = dsl.Q()
-    if (
-        request.values.get("ondemand") != "true"
-        and request.values.get("ondemand") != "ondemand"
-    ):
-        _query = _query & ~dsl.Q("match", **{"distribution.availability": "ondemand"})
+
+    #    if (
```
zzacharo, line 359: can be deleted?

cernopendata/modules/fixtures/cli.py
```diff
@@ -80,27 +85,27 @@ def _handle_record_files(record, data):
...
-            print("File index created")
             f.delete()
         elif "type" in file and file["type"] == "index.txt":
             # The txt indexes should be ignored. Just delete the file
             f.delete()
         else:
             real_files.append(file)
             try:
-                obj = ObjectVersion.create(record.bucket, filename, _file_id=f.id)
+                obj = FileObjectCold.create(record.bucket, filename, f.id)
```
zzacharo, line 102:
```suggestion
                obj = FileObjectCold.create_version(record.bucket, filename, f.id)
```

just to separate the creation of a file obj and file object version?

cernopendata/config.py
psaiz: I've put it back, since the tests were not happy. If we remove this, we break backward compatibility if a user was searching through this field

cernopendata/modules/fixtures/cli.py
psaiz: Ok, thanks!  I'll address this on https://github.com/cernopendata/cernopendata-portal/pull/137 (this PR included two commits, the one of 137 and another one based on that one).

cernopendata/api.py
```diff
@@ -174,3 +201,38 @@ def dumps(self):
     def flush(self):
         """Flushing the information into the object."""
         return self.dumps()
+
+
+class FileObjectCold(FileObject):
+    """Overwrite the fileobject to get multiple URI and availability."""
+
+    @classmethod
+    def create(self, bucket, filename, file_id):
```
psaiz, line 210:
```suggestion
    def create_version(self, bucket, filename, file_id):
```

cernopendata/api.py
```diff
@@ -174,3 +201,38 @@ def dumps(self):
     def flush(self):
         """Flushing the information into the object."""
         return self.dumps()
+
+
+class FileObjectCold(FileObject):
```
psaiz, line 206:
```suggestion
class MultiURIFileObject(FileObject):
```

cernopendata/api.py
```diff
@@ -174,3 +201,38 @@ def dumps(self):
...
+    """Overwrite the fileobject to get multiple URI and availability."""
```
psaiz, line 207: explain a bit better
```suggestion
    """Overwrite the fileobject to allow multiple URI for the same instance. For example, a file could be located both on eos disk and on tape. The class introduces as well the availability concept: if a file is available only on tape, then the file has to be requested before users can access it."""
```

zzacharo (approved):

PR #140, opened 2025-02-10
openaire: Setting up oai2d server

psaiz:


cernopendata/modules/datacite/serializers.py
```diff
@@ -24,28 +24,67 @@
...
 class DataCiteSerializer(Schema):
     """DataCite complient schema."""
 
-    identifier = fields.Method("get_identifier")
-    creators = fields.Method("get_creators")
+    doi = fields.Method("get_identifier")
+    creators = fields.Method("get_creator")
     titles = fields.Method("get_titles")
-    resourceType = fields.Method("get_resourcetype")
+    # resourceType = fields.Method("get_resourcetype")
```
zzacharo, line 55: leftover?

setup.py
```diff
@@ -69,45 +69,37 @@
...
     "invenio-mail>=2.1.1",
     # Custom Invenio `opensearch` bundle
-    "invenio-search[opensearch2]>=2.3.1,<3.0.0",
+    "invenio-search[opensearch2]>=3.0.0,<4.0.0",
     # Specific Invenio dependencies
     "invenio-xrootd==2.0.0a1",
     "xrootdpyfs==2.0.0a1",
     # Specific dependencies
-    "Flask-Markdown>=0.3.0",
+    "Flask-Markdown @git+https://git@github.com/psaiz/flask-markdown",
```
zzacharo, line 102: why this pin?

setup.py
```diff
@@ -69,45 +69,37 @@
...
+    "invenio-records-ui>=2.0.0,<3.0.0",
+    "invenio-search-ui>=4.0.0,<5.0.0",
     "invenio-records-files>=1.2.1,<3.0.0",
-    "invenio-stats>=4.3.0,<5.0.0",
+    "invenio-stats>=5.1.0,<6.0.0",
     "jupyter-client==7.1.0",
+    "invenio-jsonschemas>=2.0.0,<3.0.0",
     "pluggy==0.13.1",
+    #    "invenio-oaiserver>=3.1.0,<4.0.0",
+    "invenio-oaiserver @git+https://git@github.com/psaiz/invenio-oaiserver@component",
```
zzacharo, line 92: temporary?

cernopendata/modules/theme/static/oai2.xsl
```diff
@@ -0,0 +1,658 @@
+<?xml version="1.0" encoding="utf-8"?>
```
zzacharo, line 1: this is something each service needs to implement?

setup.py
psaiz: Yep, thanks. I've increased the required version of oaiserver

setup.py
psaiz: This is to be able to move to Flask 3. The module needed to merge a PR that someone submitted some years ago. Unfortunately, the module seems to be abandoned. I've cloned it, applied the PR, and that works

cernopendata/modules/datacite/serializers.py
psaiz: Thanks, removed

cernopendata/modules/theme/static/oai2.xsl
psaiz: I'm following the example of zenodo. Indeed, it would be better if it was provided by invenio-oaiservice. I'm happy to move it there if you prefer

setup.py
zzacharo: where do we use it? it is fine for now but maybe open an issue to see if we can replace or find another solution for this?

cernopendata/modules/theme/static/oai2.xsl
zzacharo: I am not familiar with the usage of this but if it something every instance will copy, then moving it there make sense.

zzacharo (approved):

PR #167, opened 2025-06-02
fix: The download index button only shows two options if all the file…

psaiz:
…s are online

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
@@ -46,6 +46,7 @@ function FileActionsDropdown({ file, table_type, setOpenModal, setSelectedFile, 
     }, 0);
     return () => clearTimeout(timeout);
   }, []);
+  const isOnline = Object.keys(file.availability).length === 1 && file.availability.hasOwnProperty('online');
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
```diff
@@ -46,6 +46,7 @@ function FileActionsDropdown({ file, table_type, setOpenModal, setSelectedFile,
     }, 0);
     return () => clearTimeout(timeout);
   }, []);
+  const isOnline = Object.keys(file.availability).length === 1 && file.availability.hasOwnProperty('online');
```
zzacharo, line 49: if a file has both hot and cold copies it is not marked as online?

cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
```diff
@@ -82,18 +85,20 @@ function FileActionsDropdown({ file, table_type, setOpenModal, setSelectedFile,
...
-          title="Download this file in JSON format"
-          download
-        >
-          Download json of hot files
+          Format: json
         </a>
+        { !isOnline && (
+            <a
+              className="item ui fluid button"
+              href={`${getFileUri(table_type, file.key)}?qos=hot`}
```
zzacharo, line 93: nit: I would try to pass down the values from the `ColdStorageActions` so we don't hardcode the values `hot`, `cold` etc.

zzacharo (approved):

cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
psaiz: Yes, if a file has two copies (and one of them is online), then the file is online

cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
psaiz: Good point. I'll create a ticket to change that (since that requires also a couple of changes in the python class and I don't dare to do it right now)

PR #171, opened 2025-06-17
UI: Created a new year facet

### LINKED ISSUE #139: Change `year` facet
The facet to select the year in opendata-qa has a lot of options. 

Image

It might be better to change it for something like the one from our videos colleagues:

Image

dianarand:
Closes #139

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/components/RangeAggregation.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/components/RangeAggregation.js
@@ -2,12 +2,19 @@ import _debounce from 'lodash/debounce';
 import PropTypes from "prop-types";
 import React, { useState, useMemo, useEffect, useRef, useCallback } from "react";
 import { useSelector } from "react-redux";
-import { FlexibleWidthXYPlot, VerticalRectSeries, Hint } from "react-vis";
+import {
+  ResponsiveContainer,
+  BarChart,
+  Bar,
+  XAxis,
+  YAxis,
+  Tooltip,
+  Cell,
+} from "recharts";
 import { Card } from "semantic-ui-react";
 import RangeSlider from './RangeSlider';
 import { withState } from 'react-searchkit';
 
-const HALF_BAR_WIDTH = 0.4;
 const SELECTED_COLOR = "#91d5ff";
 const DESELECTED_COLOR = "#bfbfbf";
 const HIGHLIGHT_COLOR = "#69c0ff";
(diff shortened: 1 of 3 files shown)


cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
```diff
@@ -39,9 +40,13 @@ const CODFacets = ({ aggs, updateQueryState, currentQueryState }) => {
         <div class="content"><ActiveFilters /></div>
       </div>
 
-      {aggs.map((agg) => (
-        <BucketAggregation key={agg.title} title={agg.title} agg={agg.agg} />
-      ))}
+      {aggs.map((agg) =>
+        agg.agg.aggName === 'year' ? (
```
psaiz, line 44: I was wondering: instead of doing this based on the name of the aggregation, would it be possible to make it based on the type of aggregation (term_filter or range_filter)? If so, we would also get this new facet for the number of events, which would be cool

cernopendata/modules/theme/assets/semantic-ui/js/search/components/RangeAggregation.js
```diff
@@ -0,0 +1,206 @@
...
+import { useSelector } from "react-redux";
+import { FlexibleWidthXYPlot, VerticalRectSeries, Hint } from "react-vis";
+import { Card } from "semantic-ui-react";
+import RangeSlider from './RangeSlider';
+import { withState } from 'react-searchkit';
+import { safeParse } from "../utils";
+
+const HALF_BAR_WIDTH = 0.4;
+const SELECTED_COLOR = "#91d5ff";
+const DESELECTED_COLOR = "#bfbfbf";
```
psaiz, line 13: Why are some colors defined here, and some in the css? Would it be possible to have all of them in the scss?

zzacharo (commented):
Are there any screenshots of how that looks like? Also, wouldn't be possible to extract this in invenio-search-ui or react-invenio-forms?

cernopendata/modules/theme/assets/semantic-ui/js/search/components/RangeAggregation.js
```diff
@@ -0,0 +1,206 @@
...
+
+const RangeAggregation = (props) => {
+  const { title, agg, currentQueryState } = props;
+  const resultsAggregations = useSelector(
+    (state) => state.results.data.aggregations
+  );
+  const initialBuckets = extractBuckets(resultsAggregations, agg.aggName)
+  const keys = initialBuckets.map(bucket => Number(bucket.key_as_string));
+  const min = Math.min(...keys) || 1900;
+  const max = Math.max(...keys) || 2025;
```
zzacharo, line 65: this should be configurable otherwise next year will be not taken into account....maybe you can do something like  pseudocode: `date.currentYear()`

ntarocco:
To make it re-usable, this should go to https://github.com/inveniosoftware/react-searchkit.
Moreover, I would suggest exploring the usage of some well-know library, so that UI widget can render nicely and customized.

zzacharo:
@dianarand @psaiz are you aware of https://github.com/uber/react-vis/issues/1485 . It seems that `react-vis` has blocking issues for upgrading React.... I think we should converge the discussion with what @ntarocco mentioned and align on how to move this forward :)

PR #176, opened 2025-07-03
Define validated_runs field

IssaAlBawwab:
Updates the record JSON schema to include the definition for the new top-level `validated_runs` field.

This ensures that all records populated with this new field will validate correctly.


psaiz:
The test that fails is due to the dependencies. Would you mind checking it updating `    "gevent==25.5.1",` in the line 117 of the setup.py fixes the issue?

psaiz (changes requested):

PR #178, opened 2025-07-04
Fix: files section is only shown when there are files

dianarand:


cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -429,7 +429,7 @@ <h2>Source code repository</h2>
 
 {% block files_block %}
     <div>
-        {% if record.distribution %}
+        {% if record.files %}
```
psaiz, line 432: What happens with the records that only have file indices?

dianarand: Added an additional check to make correctly show the section in this case as well.

psaiz (approved):

PR #191, opened 2025-07-28
fix: commit changes in clear-hot after all steps are completed

### LINKED ISSUE #175: cold command: clear_hot is not transactional
Take a look a the command:

cernopendata cold clear_hot <id>

It does three things:
1. It attempts to delete the local file
2. It creates an objecttag  marking the entry as deleted
3. it reevaluates the metadata of the record, and saves it on _files (or _index_files)

If, for whatever reason, the command is killed after the second step (and before the third), it leaves the repository in an inconsistent situation, where the file has been marked as deleted, the record does not know about it. Attempting to clear_hot again will give a duplicate_tag error.

Two things to do here:
1. (and most important). Check the logic, and ensure that the commit to the database is only done after 3 (and not after 2)
2. Make an option, like `-force`, that will continue even if the objecttag exists

dianarand:
Closes #175

### DIFF
--- a/cernopendata/cold_storage/catalog.py
+++ b/cernopendata/cold_storage/catalog.py
@@ -28,10 +28,9 @@ import logging
 from datetime import datetime
 
 from invenio_db import db
-from invenio_files_rest.errors import DuplicateTagError
 from invenio_files_rest.models import FileInstance, ObjectVersion, ObjectVersionTag
 from invenio_indexer.api import RecordIndexer
-from invenio_pidstore.models import PersistentIdentifier
+from sqlalchemy.exc import IntegrityError
 
 from cernopendata.api import RecordFilesWithIndex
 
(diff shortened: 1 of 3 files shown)


cernopendata/cold_storage/catalog.py
```diff
@@ -70,12 +70,14 @@ def get_files_from_record(self, record, limit=None):
...
         return files
 
-    def clear_hot(self, record, file_id):
+    def clear_hot(self, record, file_id, force):
         """Marking the hot copy as deleted."""
 
         def _clear_hot_function(version_id):
             """Create a tag for the file identifying that the copy is not available."""
             try:
+                if ObjectVersionTag.get(version_id, "hot_deleted") and force:
```
zzacharo, line 79: why do we need to add this? Wasn't enough to the execution of `ObjectVersionTag.create(version_id, "hot_deleted", str(datetime.now()))`? That should give a duplicated error no?

cernopendata/cold_storage/catalog.py
```diff
@@ -122,6 +122,7 @@ def reindex_entries(self):
                     logger.info("The second time worked!")
                 except Exception as e:
                     logger.error(f"Doing it again did not help :( {e}")
+        db.session.commit()
```
zzacharo, line 125: What is the reason of committing only to the end? There is a scenario that we might endup with indexed data while db operation crushes. Imagine that the iteration loop successfully indexes all records and the moment it exits the `while` loop the `db.session.commit()` fails...What about moving the commit after the `try/except`?

cernopendata/cold_storage/catalog.py
dianarand: I changed it so that the `IntegrityError` is caught and added a return to the handling so that if not using the --force option, it will not go to the reindex and the command will stop running.

PR #195, opened 2025-08-06
XrootD: Automatically read version from environment/container

Biscgit:


psaiz (changes requested):
I like this! Would you mind rebasing it on the new main, that have already the changes in the Dockerfile? Then, I will approve and merge this

psaiz (approved):

Biscgit:
I changed it slightly, so that when no version is specified in the container, it will install the latest python package as well (instead of the default specified)

PR #227, opened 2025-09-25
cold: add new columns to the transfer statistics

### LINKED ISSUE #208: Add new columns to the transfer statistics
Take a look at this screenshot from https://opendata.cern.ch/transfer_requests

Image

The columns of `# files`  and `size` are confusing. It  refers to the number of transfers that have been submitted. So, for records with a lot of files, that number will increase when the new transfers are submitted. 

It would be better to have the following columns:
* The number of files on the record (in the case of record 8882 for instance, that would be 1099
* The size of the files on the record (2,7Tib)
* The number of files on each state when the request was submitted. So, in this case, assuming that the record was all on cold, with one file having also a copy on hot, it would be 1 hot, 1099 cold
* The number of transfers that have been issued (the current value of `# files` )
* The size of the transfers that have been issued (the current value of `size`)

Image

dianarand:
Resolves #208

### DIFF
--- a/cernopendata/cold_storage/api.py
+++ b/cernopendata/cold_storage/api.py
@@ -187,11 +187,14 @@ class Request:
             logger.error(f"Failed to send email to {emails}: {e}")
 
     @staticmethod
-    def create(record_id, subscribers=[], file=None):
+    def create(record_id, subscribers=None, file=None, availability=None):
         """Create a new request."""
         rb = RequestMetadata(
             record_id=record_id, file=file, subscribers=subscribers or []
         )
+        if availability:
+            rb.num_hot_files = availability.get("online")
+            rb.num_cold_files = availability.get("on demand")
         db.session.add(rb)
         return rb
 
--- a/cernopendata/cold_storage/schemas.py
+++ b/cernopendata/cold_storage/schemas.py
@@ -23,8 +23,6 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 """Schemas used for the transfers."""
-
-from invenio_pidstore.models import PersistentIdentifier
 from marshmallow import Schema, fields, validate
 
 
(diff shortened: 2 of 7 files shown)


cernopendata/cold_storage/schemas.py
```diff
@@ -52,8 +50,27 @@ class TransferRequestSchema(Schema):
...
-        pid = PersistentIdentifier.query.filter_by(object_uuid=obj.record_id).first()
-        return pid.pid_value if pid else None
+        pid_mappings = self.context.get("pid_mappings", {})
+        recid = pid_mappings.get(obj.record_id)
+        return recid if recid else None
+
+    def get_num_record_files(self, obj):
+        """Get the number of files from the record data."""
+        record_data = self.context.get("records_data", {}).get(str(obj.record_id))
+        if record_data:
```
psaiz, line 67: can we catch the case where the record_data does not have the `distribution` field? It is not supposed to happen. At the same time, it would be better that the functions returns None instead of crashing

cernopendata/cold_storage/api.py
```diff
@@ -187,11 +187,13 @@ def send_email(req, emails):
             logger.error(f"Failed to send email to {emails}: {e}")
 
     @staticmethod
-    def create(record_id, subscribers):
+    def create(record_id, subscribers, availability):
```
psaiz, line 190: what about giving a default value to availability (and then maybe also to subscribers)? 
That way, the tests do not have to be modified

psaiz (changes requested):
See the three open conversations, please

cernopendata/cold_storage/schemas.py
dianarand: Added a condition for the `distribution` field existing, if not, the function returns `None`.

PR #277, opened 2025-12-18
fix: ensure invalid emails or empty strings are not added as subscribers

### LINKED ISSUE #271: Error during sending out email notifications
When a transfer request is marked as completed and email notifications are attempted to be sent, there occurs an error
`Failed to send email to ['']: {'': (501, b'5.1.3 Invalid address')}`

Ensure there is no empty string in the list of subscribers to avoid this error.

dianarand:
Resolves #271 
In addition to fixing the problem described in the issue, both backend and frontend validation for emails when subscribing were added and some minor UI/UX improvements were made.

### DIFF
--- a/cernopendata/cold_storage/api.py
+++ b/cernopendata/cold_storage/api.py
@@ -29,6 +29,7 @@ import logging
 from datetime import datetime
 from enum import Enum
 
+from email_validator import EmailNotValidError, validate_email
 from flask import current_app
 from flask_mail import Message
 from invenio_db import db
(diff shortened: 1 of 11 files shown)


cernopendata/modules/theme/assets/semantic-ui/js/utils.js
```diff
@@ -0,0 +1,6 @@
+const EMAIL_REGEX = /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+/;
```
psaiz: Same thing here. I'm afraid that this regex allows things that are not valid emails. Are there any javscripts that do the email validattion?

cernopendata/cold_storage/api.py
```diff
@@ -194,8 +195,13 @@ def create(
         record_id, subscribers=None, file=None, availability=None, distribution=None
     ):
         """Create a new request."""
+        valid_subscribers = []
+        if subscribers:
+            valid_subscribers = [
+                Request.validate_email_address(subscriber) for subscriber in subscribers
```
psaiz: What about a warning for each subscriber that is not valid?

dianarand: My idea here was that if an email is not valid, the exception is thrown and the API endpoint will return a 400 with the corresponding error message. Currently, the only place where this `Request.create()` is used, is in the `stage` function in `cernopendata/modules/records/utils.py`, where the exception is caught.

PR #290, opened 2026-03-13
Include first prototype of the release procedure.

psaiz:


cernopendata/modules/releases/views.py
```diff
@@ -0,0 +1,362 @@
...
+    if source == "file":
+        file = request.files.get("file")
+        if not file.filename.endswith(".json"):
+            return jsonify({"error": "Only JSON files are allowed"}), 400
+
+        try:
+            payload = json.load(file)
+        except Exception as e:
+            return jsonify({"error": "Invalid JSON", "details": str(e)}), 400
+        release_name = file.filename.rsplit("/", 1)[-1]
```
dianarand: Maybe some filename sanitation could be useful here?

cernopendata/modules/releases/api.py
```diff
@@ -0,0 +1,427 @@
...
+            self.change_status(ReleaseStatus.READY, current_user)
+        else:
+            self.change_status(ReleaseStatus.DRAFT, current_user)
+
+    def deploy(self, schema, current_user):
+        """Deploy the entries of a release."""
+        if not self.is_status(ReleaseStatus.DEPLOYING):
+            raise RuntimeError("Release is not READY")
+
+        for record_data in self._metadata.records:
```
dianarand: It seems to me that if an error occurs at some point during this for loop, then some of the records will have been committed, while others are not, and the release itself is not updated. Maybe we need some rollback logic here in case something like that happens? Also potentially the same issue below in the `publish()` method.

dianarand (changes requested):

cernopendata/config.py
```diff
@@ -713,3 +756,52 @@ def _query_parser_and(qstr=None):
...
+    "consumer_secret": os.environ.get(
+        "CERNOPENDATA_CERN_APP_CREDENTIALS", "<CHANGEME>"
+    ),
+}
+
+accounts_rest.default_user_payload = user_payload_with_cern_roles
+handlers["signup_handler"]["info"] = user_info_with_cern_roles
+
+# OAUTHCLIENT_SETTINGS_TEMPLATE = 'invenio_theme/page_settings.html'
+COMMUNITIES_CUSTOM_FIELDS = None
```
zzacharo, line 811: by default this is set to `[]`, see https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/config.py#L252

scripts/create-instance.sh
```diff
@@ -39,4 +39,10 @@ cernopendata collect -v
 # be overwritten with the ones from opendata.
 for d in $(find /code/cernopendata/modules/theme/static/ -type 'f' | sed 's/\/code\/cernopendata\/modules\/theme\/static//' ) ; do cp --remove-destination /code/cernopendata/modules/theme/static${d}  ${INVENIO_INSTANCE_PATH}/static${d}; done;
 
-cernopendata webpack clean buildall
+cernopendata webpack clean
+# TODO THIS IS A HACK. If we don't create the directory by hand, we get:
+# ERROR in ./js/invenio_communities/community/new.js 207:19-76
+# 31 110.8 Module not found: Error: Can't resolve '@templates/custom_fields' in '/opt/invenio/var/instance/assets/js/invenio_communities/community'
```
zzacharo: The passed prop to the UI is [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/views/communities.py#L261), can you check that this is indeed empty?

It is used in the frontend [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/CommunityProfileForm.js#L594). Does the error happen because it is just being imported? If so, we can probably change the import to a dynamic one inside that if block.

cernopendata/modules/releases/validations/expand_files.py
```diff
@@ -0,0 +1,105 @@
...
+    def validate(self, release):
+        """Check if there are any directories as input for a record."""
+        errors = []
+        for record in release.records:
+            if "files" not in record:
+                continue
+
+            for file in record["files"]:
+                if "uri" in file and file["uri"].endswith("*"):
+                    errors.append(f"The record {i} has a path like {file['uri']}")
```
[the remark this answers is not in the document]
psaiz:
```suggestion
                    errors.append(f"The record has a path like {file['uri']}")
```

cernopendata/modules/releases/validations/files.py
```diff
@@ -0,0 +1,61 @@
...
+                        errors.append(
+                            f"Entry {i + 1}, file {j + 1}: Missing size/checksum"
+                        )
+        return errors
+
+    def _get_entry_details(self, ctx, url):
+        """Given a url, return if the entry is a directory or a file. In case of file, return size anc checksum."""
+        st = ctx.stat(url)
+
+        if st.st_mode & 0o40000:  # POSIX directory flag
```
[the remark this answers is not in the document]
psaiz:
```suggestion
        if stat.S_ISDIR(st.st_mode):
```

cernopendata/modules/releases/validations/files.py
```diff
@@ -0,0 +1,61 @@
+"""Validation process."""
+
+import gfal2
```
psaiz, line 5:
```suggestion
import gfal2
import stat
```

cernopendata/modules/releases/models.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class ReleaseMetadata(db.Model):
+    """Release model."""
+
+    __tablename__ = "releases_metadata"
+
+    BULK_IMMUTABLE_FIELDS = ["recid", "title", "DOI"]
+    status = db.Column(
+        db.String(20),
```
psaiz:
```suggestion
        db.Enum(ReleaseStatus),
```

cernopendata/modules/releases/views.py
```diff
@@ -0,0 +1,362 @@
...
+        abort(403)
+    source = request.form.get("source")
+
+    if source == "file":
+        file = request.files.get("file")
+        if not file.filename.endswith(".json"):
+            return jsonify({"error": "Only JSON files are allowed"}), 400
+
+        try:
+            payload = json.load(file)
```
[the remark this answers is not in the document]
psaiz: I'm hoping that we don't have such big files. Releases could be split over multiple files (see for instance how it was done for the skeletons)

cernopendata/modules/releases/views.py
psaiz: The user should be able to modify the release name. This is just to give a default value at the beginning

cernopendata/modules/releases/views.py
```diff
@@ -0,0 +1,362 @@
...
+    # Get JSON string from form
+    import sys
+
+    print("HELLO", file=sys.stderr)
+    data = request.get_json()
+    if not data["records"]:
+        abort(400, "No records provided")
+
+    if not isinstance(data["records"], list):
+        raise ValueError("Records must be a list")
```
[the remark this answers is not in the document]
psaiz:
```suggestion
        flash("Records must be a list", "error")
        abort(400, "Records must be a list")
```

cernopendata/modules/releases/api.py
psaiz: Note that the session commit is only at the bottom, and the record is not committed until the session commit happens

cernopendata/modules/releases/views.py
```diff
@@ -0,0 +1,362 @@
...
+    template_folder="templates",
+    static_folder="static",
+)
+
+
+@blueprint.route("/releases/api/list/<experiment>", methods=["GET"])
+@login_required
+def list_releases(experiment=None):
+    """Return list of releases."""
+    if not Release.validate_experiment(experiment):
```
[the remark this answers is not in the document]
psaiz: Moved, thanks

dianarand (approved):

PR #297, opened 2026-04-13
Record authentication

dianarand:
Resolves #291

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -346,47 +346,57 @@ RECORDS_UI_EXPORT_FORMATS = dict(
     ),
 )
 
+_RECORD_PERMISSION_FACTORY = (
+    "cernopendata.modules.records.permissions:record_read_permission_factory"
+)
+
 RECORDS_UI_ENDPOINTS = dict(
     recid=dict(
         pid_type="recid",
         route="/record/<pid_value>",
-        permission_factory_imp=None,
+        permission_factory_imp=_RECORD_PERMISSION_FACTORY,
         record_class="cernopendata.api:RecordFilesWithIndex",
         view_imp="cernopendata.modules.records.utils:record_metadata_view",
     ),
     recid_files=dict(
         pid_type="recid",
         route="/record/<pid_value>/files/<path:filename>",
+        permission_factory_imp=_RECORD_PERMISSION_FACTORY,
         view_imp="cernopendata.modules.records.utils:file_download_ui",
         record_class="cernopendata.api:RecordFilesWithIndex",
     ),
     recid_file_index=dict(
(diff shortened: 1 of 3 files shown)


cernopendata/modules/records/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+            experiment, release_id = prerelease.split("/", 1)
+            release_id = int(release_id)
+        except (ValueError, AttributeError):
+            return True
+
+        release = ReleaseMetadata.query.filter_by(
+            id=release_id, experiment=experiment
+        ).first()
+
+        if release is None or release.status == ReleaseStatus.PUBLISHED.value:
```
psaiz: We don't need this check. If the release is published, the record should not be in 'prerelease'

cernopendata/modules/records/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+    def can(self):
+        prerelease = record.get("prerelease")
+        if not prerelease:
+            return True
+
+        try:
+            experiment, release_id = prerelease.split("/", 1)
+            release_id = int(release_id)
+        except (ValueError, AttributeError):
+            return True
```
psaiz: shouldn't this return False? If we get to this except, there is an error in our logic, so it should also print a message in the log

cernopendata/config.py
```diff
@@ -341,54 +341,65 @@
...
+_RECORD_PERMISSION_FACTORY = (
+    "cernopendata.modules.records.permissions:record_read_permission_factory"
+)
+
 RECORDS_UI_ENDPOINTS = dict(
     recid=dict(
         pid_type="recid",
         route="/record/<pid_value>",
-        permission_factory_imp=None,
+        permission_factory_imp=_RECORD_PERMISSION_FACTORY,
```
psaiz, line 357: Can the same method be used for the `docs`? For the glossary terms it is not needed

psaiz (changes requested):
Nice! Could you please add a couple of tests to make sure that future commits do not break this funcionality?

cernopendata/modules/records/permissions.py
dianarand: Thanks, removed it

cernopendata/modules/records/permissions.py
dianarand: Thanks, added logging and changed it to return False

cernopendata/config.py
dianarand: It seems to me that it should be possible to use it, so I added it. I will test it out properly once I've added the functionality for managing documents with the releases.

psaiz (approved):

PR #307, opened 2026-04-30
curation: add management of documents

dianarand:
Resolves #292 and additionally contains UI improvements for the release details view

### DIFF
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -29,6 +29,7 @@ from copy import deepcopy
 from datetime import datetime
 
 from deepdiff import DeepDiff
+from flask import current_app
 from invenio_db import db
 from invenio_indexer.api import RecordIndexer
 from invenio_pidstore.models import PersistentIdentifier
@@ -37,8 +38,11 @@ from sqlalchemy.orm import selectinload
 from sqlalchemy.orm.attributes import flag_modified
 
 from cernopendata.modules.fixtures.cli import (
+    create_doc,
     create_record,
+    delete_doc_or_glossary,
     delete_record,
+    update_doc_or_glossary,
     update_record,
 )
 
@@ -78,6 +82,16 @@ class ReleaseValidation:
         return self.validator.optional
 
     @property
+    def is_document_validation(self):
+        """True if this validation only applies to documents."""
+        return self.validator.applies_to == {"documents"}
+
+    @property
+    def is_record_validation(self):
+        """True if this validation only applies to records."""
+        return self.validator.applies_to == {"records"}
+
+    @property
     def name(self):
         """Name of the validation."""
         return self._metadata.name
@@ -107,6 +121,8 @@ class ReleaseValidation:
 
     def validate(self):
         """Run the validation."""
+        if not self.validator:
+            return []
         return self.validator.validate(self._metadata.release)
 
     def fix(self):
@@ -138,11 +154,23 @@ class ReleaseValidation:
 class Release:
     """Class for the release."""
 
-    record_schema = "local://records/record-v1.0.0.json"
-
     def __init__(self, metadata):
         """Initialize the object."""
         self._metadata = metadata
+
+    @staticmethod
+    def _record_schema_url():
+        """Return the JSON schema URL for records."""
+        return current_app.extensions["invenio-jsonschemas"].path_to_url(
+            "records/record-v1.0.0.json"
+        )
+
+    @staticmethod
+    def _document_schema_url():
+        """Return the JSON schema URL for documents."""
+        return current_app.extensions["invenio-jsonschemas"].path_to_url(
+            "records/docs-v1.0.0.json"
+        )
 
     @property
     def status(self):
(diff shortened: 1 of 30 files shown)


cernopendata/modules/releases/validations/doc_schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+import json
+
+from flask import current_app
+from jsonschema import Draft4Validator
+
+from .base import Validation
+
+
+class ValidDocSchema(Validation):
```
psaiz: I suppose that the only difference between this validation and the one for the records is the schema. What about creating a base class (similar to expected_fields), and then having a doc_schema and rec_schema inheriting from it?

cernopendata/modules/releases/validations/expected_fields.py
```diff
@@ -30,6 +30,7 @@ class ExpectedFieldsValidation(Validation):
 
     expected_fields = {}
     expected_record_fields = {}
+    expected_doc_fields = {}
```
psaiz: brainstorming: would it be better to have:
expected_fields = {} //this would be common for records and docs
applies_to  = ["records", "docs"] //or only one of the two

cernopendata/modules/releases/validations/slug.py
```diff
@@ -0,0 +1,77 @@
+"""Validation process."""
+
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from ..models import ReleaseStatus
+from .base import Validation
+
+
+class ValidSlug(Validation):
+    """Check the slugs of the documents in a release."""
```
psaiz, line 7: This one should be very similar to the recid. Any chance of making them compatible?

cernopendata/modules/releases/api.py
```diff
@@ -77,6 +81,16 @@ def optional(self):
         """Boolean to check if the validation is optional."""
         return self.validator.optional
 
+    @property
+    def is_document_validation(self):
+        """True if this validation only applies to documents."""
```
psaiz, line 86: If a validation applies to both, are both true? Then the `only` part is not correct

cernopendata/modules/releases/api.py
```diff
@@ -316,13 +374,21 @@ def validate(self, current_user):
...
+            self.change_status(ReleaseStatus.DRAFT, current_user)
+            db.session.commit()
+            raise RuntimeError("Release has validation errors and cannot be staged")
+
+        schema = current_app.extensions["invenio-jsonschemas"].path_to_url(
+            self.record_schema_path
+        )
         for record_data in self._metadata.records:
-            record_data["$schema"] = Release.record_schema
+            record_data["$schema"] = schema
```
psaiz: I wonder if it would be better to populate this field during the release creation (both for documents and recods)

cernopendata/modules/releases/views.py
```diff
@@ -317,6 +340,92 @@ def publish(experiment, release_id):
...
+                jsonify(
+                    {
+                        "error": f"Document '{doc.get('slug', '?')}' has a filename pointer "
+                        "in body.content. Inline the markdown text before uploading."
+                    }
+                ),
+                400,
+            )
+
+    release = _get_release(experiment, release_id)
```
psaiz: This is the one that checks the permissions, isn't it? Shouldn't it be done before the rest of the logic?

cernopendata/modules/releases/views.py
```diff
@@ -317,6 +340,92 @@ def publish(experiment, release_id):
...
+        {"status": "ok", "num_docs": release._metadata.num_docs, "documents": docs}
+    )
+
+
+@blueprint.route(
+    "/releases/<experiment>/<int:release_id>/update_document",
+    methods=["POST"],
+)
+@login_required
+def update_document(experiment, release_id):
```
psaiz: If this updates a single document, what about putting the slug in the URL?

cernopendata/modules/theme/assets/semantic-ui/js/curate/DocumentsTable.js
```diff
@@ -0,0 +1,392 @@
...
+  Modal,
+  Form,
+  TextArea,
+  Tab,
+  Radio,
+  Input,
+  Message,
+} from "semantic-ui-react";
+
+export default function DocumentsTable({
```
psaiz, line 16: brainstorming: do we need completely different table, or could we reuse the same component for records and documents?

cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss
```diff
@@ -556,3 +556,191 @@ a {
...
+}
+
+.release-step-substatus {
+    font-weight: normal;
+    font-size: 0.85em;
+}
+
+.doc-hint {
+    margin-top: 0.4em;
+    color: rgba(0, 0, 0, 0.5);
```
psaiz: I wonder if this could be done with semantic-ui instead of css

psaiz (changes requested):

cernopendata/modules/releases/validations/doc_schema.py
dianarand: Good idea, thanks. I made `schema.py` into the base class, and in `doc_schema.py‎` and `rec_schema.py‎` are the inheriting classes.

cernopendata/modules/releases/validations/slug.py
dianarand: I created a base class for these in `pid.py` and there are now `recid.py` and `slug.py` inheriting from it.

cernopendata/modules/releases/api.py
dianarand: I fixed this so that it is `True` only if it applies to documents.

cernopendata/modules/releases/views.py
dianarand: Thanks for noticing, I moved it accordingly.

cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss
dianarand: I now removed some of the custom CSS that was possible to do with semantic-ui, or was just too minor to be useful.

cernopendata/modules/releases/api.py
dianarand: I updated this so that it is done during release creation or when adding/updating documents/records.

cernopendata/modules/releases/validations/expected_fields.py
dianarand: I improved this by only having `expected_fields` (common for records and docs)

cernopendata/modules/releases/views.py
dianarand: Thanks, done

cernopendata/modules/theme/assets/semantic-ui/js/curate/DocumentsTable.js
dianarand: There are some parts that could be reused, but there are also a lot of differences, which is why I chose to create a separate one for documents. For example, the functionalities of adding and editing and the related modals are different, and these take up a large part of the components. I think these might be more easily readable as separate components.

dianarand: Created separate issue #309 to improve this

psaiz (approved):

PR #314, opened 2026-05-07
feat(releases): add image upload for documents

### LINKED ISSUE #305: Interface to upload the images needed by the documents
Follow up for #292 

Create the functionalities needed to support the upload of images for the documents. This includes:
- creating a new persistent volume claim to store the images
- updating nginx configuration to serve the uploaded images
- adding endpoints to upload and delete images
- updating staging/rollbacking/publishing releases to handle uploaded images
- updating the user interface to handle uploading images and viewing them
- adding a validation to ensure the image links in the document point to existing images

dianarand:
Resolves #305

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -675,6 +675,10 @@ CERNOPENDATA_MAX_DOWNLOAD_SIZE = os.environ.get(
 )
 #: Make download availalbe for Root files
 CERNOPENDATA_DISABLE_DOWNLOADS = os.environ.get("CERNOPENDATA_DISABLE_DOWNLOADS", False)
+#: Document image storage
+CERNOPENDATA_IMAGES_PATH = os.environ.get(
+    "CERNOPENDATA_IMAGES_PATH", "/opt/invenio/var/instance/static/upload"
+)
 # Search
 # ======
 #: Default OpenSearch document type.
--- a/cernopendata/modules/releases/validations/image_references.py
+++ b/cernopendata/modules/releases/validations/image_references.py
@@ -1,0 +1,88 @@
+"""Validation that all images referenced in document bodies are available."""
+
+import re
+from pathlib import Path
+from urllib.parse import unquote
+
+from flask import current_app
+
+import cernopendata
+
+from .base import Validation
+
+
+class ValidImageReferences(Validation):
+    """Check that images referenced in a markdown body are available."""
+
+    name = "Valid image references"
+    error_message = (
+        "Some images referenced in document bodies are not available. "
+        "Upload the missing images or correct the path."
+    )
+    applies_to = {"documents"}
+
+    def validate(self, release):
+        """Walk every markdown body and check each image reference."""
+        errors = []
+        uploaded_image_storage = Path(
+            current_app.config["CERNOPENDATA_IMAGES_PATH"]
+        ).resolve()
+        repo_images = (
+            Path(cernopendata.__file__).parent / "modules" / "theme" / "static" / "docs"
+        ).resolve()
+        for i, doc in enumerate(release.documents or []):
+            body = doc.get("body") or {}
+            if body.get("format") != "md":
+                continue
+            content = body.get("content")
+            if not isinstance(content, str) or not content:
+                continue
+            for url in self._extract_image_urls(content):
+                error = self._check_url(url, uploaded_image_storage, repo_images)
+                if error:
+                    errors.append(f"Document {i + 1}: {error}")
+        return errors
+
+    @staticmethod
+    def _extract_image_urls(content):
+        """Return image URLs referenced in the content."""
+        html_urls = re.findall(
+            r"""<img\b[^>]*?\bsrc\s*=\s*["']([^"']+)["']""",
+            content,
+            re.IGNORECASE,
+        )
+        md_urls = re.findall(r"!\[[^\]]*\]\(\s*([^)\s]+)", content)
+        image_urls = []
+        for url in (*html_urls, *md_urls):
+            url = url.strip()
+            if url and url not in image_urls:
+                image_urls.append(url)
+        return image_urls
+
+    @staticmethod
+    def _check_url(url, uploaded_image_storage, repo_images):
+        """Return an error string if the URL points to a missing local image."""
+        cleaned_url = url.split("#", 1)[0].split("?", 1)[0]
+        for prefix, root in (
+            ("/static/upload/", uploaded_image_storage),
+            ("/static/docs/", repo_images),
+        ):
+            if not cleaned_url.startswith(prefix):
+                continue
+            relative_path = unquote(cleaned_url[len(prefix):])  # fmt: skip
+            if not relative_path:
+                return f"image referenced in body has invalid path: {url}"
+            full_path = (root / relative_path).resolve()
+            try:
+                full_path.relative_to(root)
+            except ValueError:
+                return f"image referenced in body has invalid path: {url}"
+            if not full_path.is_file():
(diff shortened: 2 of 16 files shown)


cernopendata/modules/releases/validations/image_references.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ValidImageReferences(Validation):
+    """Check that images referenced in a markdown body are available."""
+
+    name = "Valid image references"
+    error_message = (
+        "Some images referenced in document bodies are not available. "
+        "Upload the missing images or correct the path."
+    )
+    applies_to = {"documents"}
```
psaiz, line 22: What about making it also for the records? See for instance https://opendata.cern/record/12350

cernopendata/modules/releases/views.py
```diff
@@ -400,6 +407,230 @@ def add_documents(experiment, release_id):
...
 
 
+@blueprint.route(
+    "/releases/<experiment>/<int:release_id>/upload_image",
+    methods=["POST"],
+)
+@login_required
+def upload_image(experiment, release_id):
+    """Upload one or more images attached to a parent document."""
+    parent_slug = request.form.get("parent_slug")
```
psaiz, line 417: It would be nice if we do it either from slug or from recid, so that it could be used for records as well

cernopendata/modules/releases/views.py
```diff
@@ -400,6 +407,230 @@ def add_documents(experiment, release_id):
...
+    target_dir = (images_root / parent_slug).resolve()
+    try:
+        target_dir.relative_to(images_root)
+    except ValueError:
+        abort(400, "Invalid parent_slug")
+    target_dir.mkdir(parents=True, exist_ok=True)
+
+    images = []
+    for file in files:
+        filename = secure_filename(file.filename or "").lower()
```
psaiz, line 448: If several images have the same name, would they overwrite each other?

psaiz (changes requested):

cernopendata/modules/releases/views.py
dianarand: Thanks for pointing this out, I made some changes so that overwriting wouldn't happen.

cernopendata/modules/releases/validations/image_references.py
dianarand: This example has only images that are external and for now I didn't find any records with images which weren't external. This validator does not validate external images, only the ones that are uploaded via this feature or added via the [opendata.cern.ch](https://github.com/cernopendata/opendata.cern.ch) repository.

If we do want to have image uploads for records as well, it will require some additional changes, not only to the validator, but to also all of the endpoints related to images, as well as some frontend changes, so it might be worth implementing in a separate PR.

cernopendata/modules/releases/views.py
dianarand: Relevant reply in the other comment

docker-compose.yml
```diff
@@ -53,6 +53,7 @@ services:
     environment: *common-variables
     volumes:
       - web_data:/opt/invenio/var/instance/static
+      - images_data:/images
```
psaiz: I don't like so much that this directory is at the top level. 
What about putting it inside /opt/invenio/var/instance/static/upload?

tests/releases/test_validator_image_references.py
```diff
@@ -0,0 +1,149 @@
...
+        '<img src="/images/a/one.png">\n'
+        "![](/images/a/two.png)\n"
+        '<IMG SRC="/images/a/three.png">'
+    )
+    urls = ValidImageReferences._extract_image_urls(content)
+    assert set(urls) == {
+        "/images/a/one.png",
+        "/images/a/two.png",
+        "/images/a/three.png",
+    }
```
psaiz, line 156: I'm quite worried about uploading an image that would overwrite an existing image. Can we have a test to ensure that this does not happen?

nginx/localhost.conf
```diff
@@ -34,9 +34,9 @@ server {
     error_log /var/log/nginx/error.log error;
 
     # Allow only GET and HEAD requests methods. This disables unnecessary
-    # methods such as CONNECT, DELETE, OPTIONS, PATCH, POST, PUT, TRACE for
+    # methods such as CONNECT, OPTIONS, PATCH, POST, PUT, TRACE for
     # improved safety.
-    if ($request_method !~ ^(GET|HEAD|POST|PUT)) {
+    if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE)) {
```
psaiz: This if is becoming less and less useful. What do you think about having it inside some of the locations? We could allow the DELETE only in the new upload area.

cernopendata/config.py
```diff
@@ -675,6 +675,8 @@ def _query_parser_and(qstr=None):
 )
 #: Make download availalbe for Root files
 CERNOPENDATA_DISABLE_DOWNLOADS = os.environ.get("CERNOPENDATA_DISABLE_DOWNLOADS", False)
+#: Document image storage
+CERNOPENDATA_IMAGES_PATH = os.environ.get("CERNOPENDATA_IMAGES_PATH", "/images")
```
psaiz: (mentioned in another comment, I would prefer a different default directory instead of   /images  )

psaiz (changes requested):

tests/releases/test_validator_image_references.py
dianarand: There are tests for this in `tests/releases/test_views.py`: 
- `test_upload_image_rejects_collision_with_existing_file`
- `test_upload_image_rejects_collision_within_batch`

docker-compose.yml
dianarand: I changed it to `/opt/invenio/var/instance/static/upload`

cernopendata/config.py
dianarand: Changed as requested, mentioned in other comment

nginx/localhost.conf
dianarand: I removed this and added specific limits for locations and allowing `DELETE` only for image uploads.

nginx/localhost.conf
psaiz:
```suggestion
```

nginx/localhost.conf
```diff
@@ -90,7 +83,19 @@ server {
         uwsgi_ignore_headers Set-Cookie;
     }
 
+    # Image management API.
+    location ~ ^/releases/[^/]+/[0-9]+/images {
```
psaiz:
```suggestion
    location ~ ^/releases/ {
```

nginx/localhost.conf
```diff
@@ -90,7 +83,19 @@ server {
...
+        proxy_pass http://web-app;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        client_max_body_size 100M;
+    }
+
     location / {
+        limit_except GET HEAD POST PUT { deny all; }
```
psaiz:
```suggestion
        limit_except GET HEAD { deny all; }
```

nginx/localhost.conf
```diff
@@ -90,7 +83,19 @@ server {
         uwsgi_ignore_headers Set-Cookie;
     }
 
+    # Image management API.
```
psaiz:
```suggestion
    # Releases
```

psaiz (approved):

PR #319, opened 2026-05-15
feat(releases): add multiple documents via folder upload

### LINKED ISSUE #303: Adding multiple documents to a release
Follow up for #292 

In some cases, a large amount of documents might be added in one release. With the current implementation, each file has to be added separately. Create an improved user experience for adding a large amount of documents at once. A possible solution could be a folder upload.

These documents might also be interlinked. Add functionality to manage these links easily.

For example, see
https://github.com/cernopendata/opendata.cern.ch/tree/master/data/docs/lhcb-stripping
https://opendata.cern.ch/docs/stripping21r1-bhadron-b2d0pid2kspi0hhddmergedbeauty2charmline

dianarand:
Resolves #303

### DIFF
--- a/cernopendata/modules/releases/validations/base.py
+++ b/cernopendata/modules/releases/validations/base.py
@@ -50,7 +50,6 @@ class Validation:
         """Optional fix method."""
         raise NotImplementedError
 
-    @property
-    def fixable(self):
-        """Check if a validation has a fix method."""
+    def fixable(self, release=None):
+        """Check whether this validation has fixes to apply to the given release."""
         return self.fix.__func__ is not Validation.fix
--- a/cernopendata/modules/releases/validations/doc_links.py
+++ b/cernopendata/modules/releases/validations/doc_links.py
@@ -1,0 +1,137 @@
+"""Validation that all /docs/ links in document bodies resolve."""
+
+import re
+
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from .base import Validation
+
+MARKDOWN_LINK = re.compile(r"(?<!!)\[[^\]]*\]\(([^)]+)\)")
+SLUG_CHARS = re.compile(r"^[A-Za-z0-9_-]+$")
+
+
+class ValidDocLinks(Validation):
+    """Check that document links in markdown bodies resolve."""
+
+    name = "Valid document links"
+    error_message = "Some document links do not resolve."
+    applies_to = {"documents"}
+
+    def validate(self, release):
+        """Check each doc link resolves: it can be a slug in the current release or a registered PID."""
+        release_slugs = self._get_release_slugs(release)
+        errors = []
+        for i, doc in enumerate(release.documents or []):
+            content = self._get_content(doc)
+            for slug in self._extract_absolute_slugs(content):
+                if slug in release_slugs or self._is_registered_pid(slug):
+                    continue
+                errors.append(f"Document {i + 1}: link to /docs/{slug} not found")
+            for slug in self._extract_bare_slugs(content):
+                # Bare slugs need addressing: fixable if the target is known, broken otherwise.
+                if slug in release_slugs or self._is_registered_pid(slug):
+                    errors.append(
+                        f"Document {i + 1}: bare-slug link '{slug}' should be /docs/{slug}"
+                    )
+                else:
+                    errors.append(
+                        f"Document {i + 1}: link '{slug}' does not resolve to a known document"
+                    )
+        return errors
+
+    def fixable(self, release):
+        """Return True if there are any auto-fixable links."""
+        release_slugs = self._get_release_slugs(release)
+        for doc in release.documents or []:
+            content = self._get_content(doc)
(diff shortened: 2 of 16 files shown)


cernopendata/modules/releases/validations/doc_links.py
```diff
@@ -0,0 +1,137 @@
...
+            if not isinstance(content, str) or not content:
+                continue
+            for slug in self._extract_doc_link_slugs(content):
+                if slug in release_slugs or self._is_registered_pid(slug):
+                    continue
+                errors.append(f"Document {i + 1}: link to /docs/{slug} not found")
+            for slug in self._extract_bare_slugs(content):
+                if slug in release_slugs or self._is_registered_pid(slug):
+                    errors.append(
+                        f"Document {i + 1}: bare-slug link '{slug}' should be /docs/{slug}"
```
psaiz, line 34: This looks strange. Both sides of the if/else are error. Is it intended?

cernopendata/modules/releases/validations/base.py
```diff
@@ -54,3 +54,7 @@ def fix(self, release):
     def fixable(self):
         """Check if a validation has a fix method."""
         return self.fix.__func__ is not Validation.fix
+
+    def has_pending_fixes(self, release):
```
psaiz: If I get it right, the idea of this method is to be able to pass the release, so that the validator checks for this particular case, isn't it? I wonder if, instead of creating a new method, it would be better to change the signature of the fixable

cernopendata/modules/releases/validations/doc_links.py
```diff
@@ -0,0 +1,137 @@
...
+                if slug in release_slugs or self._is_registered_pid(slug):
+                    return True
+        return False
+
+    def fix(self, release):
+        """Rewrite matched bare-slug links to /docs/<slug> in each markdown body."""
+        release_slugs = {
+            doc.get("slug") for doc in (release.documents or []) if doc.get("slug")
+        }
+        for doc in release.documents or []:
```
psaiz, line 55: Lines 27-33, 54-60 and 70-77 seem quite similar. What about creating a private method that finds the slugs, and the three methods use it?

cernopendata/modules/releases/validations/doc_links.py
```diff
@@ -0,0 +1,137 @@
...
+            }
+            if not known_bare:
+                continue
+            body["content"] = self._rewrite_bare_slugs(content, known_bare)
+        return []
+
+    @classmethod
+    def _extract_doc_link_slugs(cls, content):
+        """Return unique /docs/<slug> slugs referenced in markdown links."""
+        slugs = []
```
psaiz, line 88: Do I get it that the extract_doc_links are the correct and the bare_slugs are the incorrect?

cernopendata/modules/releases/validations/doc_links.py
```diff
@@ -0,0 +1,137 @@
...
+                return match.group(0)
+            suffix = raw_target[len(slug_part):]  # fmt: skip
+            return f"[{link_text}](/docs/{slug_part}{suffix})"
+
+        return re.sub(r"(?<!!)\[([^\]]*)\]\(([^)]+)\)", rewrite, content)
+
+    @staticmethod
+    def _is_registered_pid(slug):
+        """Return True if slug is a registered docid PID."""
+        return (
```
psaiz, line 130: Imagine that a document has a slug pointing to something in the same release. That one will not have the PID registered yet, so it will fail, won't it?

psaiz (changes requested):

cernopendata/modules/releases/validations/doc_links.py
dianarand: Yes, the goal was to create different error messages depending whether the auto-fixable or not. I added a comment here to hopefully make it more clear.

cernopendata/modules/releases/validations/base.py
dianarand: Thanks, good point. I removed the method and changed `fixable` instead.

cernopendata/modules/releases/validations/doc_links.py
dianarand: Thanks,  I refactored this a little bit to avoid code repetition.

cernopendata/modules/releases/validations/doc_links.py
dianarand: `_extract_doc_link_slugs` extracts links that are already using the absolute path, while `_extract_bare_slugs` extracts links that are just a bare slug that should be fixed. I renamed `_extract_doc_link_slugs` to `_extract_absolute_slugs` and updated the docstrings so hopefully it is more clear.

cernopendata/modules/releases/validations/doc_links.py
dianarand: This method would return `False`, but before even calling this method, we already check whether the slug points to something in the same release.

psaiz (approved):

PR #325, opened 2026-05-20
refactor(releases): improve error handling and error messages

### LINKED ISSUE #323: Releases: Better error handling and messages
Several routes in the releases module let predictable failures propagate as uncaught exceptions, instead of catching them and returning a useful message. For example, creating a release from an invalid URL gives `requests.exceptions.HTTPError: 404 Client Error: Not Found for url: …` on the internal error page. This would be displayed to the user in production as an "Internal Server Error" page with no information about what went wrong or how to recover.

There are a few places where similar errors can potentially arise. Verify and fix as needed:

1. `release_upload` (URL source) does not catch HTTP or JSON errors (such as the error mentioned above). 
2. `add_documents` and `add_records` catch the HTTP fetch but call `resp.json()` outside the try block, so a 200 response with non-JSON body could still crash.
3. Endpoints related to images do not contain any handling for `OSError` / `PermissionError` / `FileNotFoundError` in case of broken paths.
4. `Release.validate()` calls each validator's `validate()` method directly, so if a single validator raises an error, every route that triggers validation returns 500.

See also whether there are any other missing error handlings to fix. 

Ensure that the React components consuming the existing `jsonify({"error": …})` responses actually display the message to the user. The user should see useful error messages in most predictable cases, avoiding the "Internal Server Error" page as much as possible.

dianarand:
Resolves #323

### DIFF
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -328,7 +328,12 @@ class Release:
             except ValueError:
                 continue
             if slug_path.is_dir():
-                shutil.rmtree(slug_path)
+                try:
+                    shutil.rmtree(slug_path)
+                except OSError:
+                    current_app.logger.warning(
+                        f"Could not delete images directory {slug_path}"
+                    )
 
     def delete(self):
         """Delete a release."""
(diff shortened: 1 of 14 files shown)


cernopendata/modules/releases/api.py
```diff
@@ -328,7 +328,12 @@ def _delete_documents_images(self):
             except ValueError:
                 continue
             if slug_path.is_dir():
-                shutil.rmtree(slug_path)
+                try:
+                    shutil.rmtree(slug_path)
```
psaiz: I'm worried about this. We are removing things from the disk based on the json description of the release, and the users can modify that freely, isn't it? Couldn't they put the name of a directory that was different from the one they uploaded?

dianarand: Good point. Maybe we could have the images stored in subfolders per release, something like `/static/upload/{release_id}/{parent_slug}/{filename}`. Then it becomes easy to restrict deleting to only within the same release. Or would you prefer a different approach?

dianarand: Implemented the subfolders per release as discussed.

psaiz (approved):

PR #327, opened 2026-05-22
feat(version): proof of concept of adding version control to the records

psaiz:


cernopendata/modules/records/api.py
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from invenio_records import Record
+from copy import deepcopy
```
dianarand, line 21: This import does not seem to be in use.
```suggestion
```

cernopendata/modules/records/api.py
```diff
@@ -0,0 +1,59 @@
...
+        data["_id_field"] = id_field
+        data["_versions"].setdefault("index", 1)
+        data["_versions"].setdefault("is_latest", True)
+
+        return super().create(data, id_=id_, **kwargs)
+
+    @property
+    def conceptrecid(self):
+        """Value of the concept record. This is the meta record, pointing to the latest version."""
+        return self.get("_conceptrecid")
```
dianarand, line 44: `_conceptrecid` values doesn't seem to be actually set anywhere.

I fixed it in my [work-in-progress commit](https://github.com/cernopendata/cernopendata-portal/commit/606a39424e0e18fc9a86168222b91e44942e94c7#diff-457f880e95ca913f0f3eab70a2730baff567ca809ca06023a8b379ffd5e16376) by renaming `_concept_parent` to `_conceptrecid`, but maybe you had something different in mind?

cernopendata/config.py
```diff
@@ -446,6 +446,9 @@ def _query_parser_and(qstr=None):
         )
     else:
         _query = dsl.Q()
+
+    # Exclude the old versions
+    _query = _query & ~dsl.Q("term", **{"_versions.is_latest": False})
```
dianarand, line 451: I think the test in `test_cernopendata_query_parser.py` needs to be updated to not fail with this change

cernopendata/modules/fixtures/cli.py
```diff
@@ -332,30 +396,36 @@ def _process_fixture_files(
...
+                    if not record:
+                        return statistics
                     action = "inserted"
                 try:
                     if record:
                         record.commit()
                     db.session.commit()
                     statistics[action] += 1
                 except Exception as e:
+                    logger.info(record)
```
dianarand, line 409:
```suggestion
```

cernopendata/modules/fixtures/cli.py
```diff
@@ -262,6 +282,28 @@ def _get_list_of_fixture_files(files, type, logger):
...
+    diff = DeepDiff(
+        data,
+        record.dumps(),
+        exclude_regex_paths=[r".*\['_.*'\]", "availability", "files"],
+    )
+    if not diff:
+        return True, None, None
+    data["_concept_parent"] = record["_concept_parent"]
+    data["_versions"] = {"index": record["_versions"]["index"] + 1, "is_latest": True}
+    record["_versions"]["is_latest"] = False
```
dianarand, line 297: The current version is set to `is_latest=False` before creating the new version. 
Maybe there should be a rollback for this in case something fails.

cernopendata/modules/records/minters/recid.py
```diff
@@ -1,30 +1,11 @@
...
-from invenio_oaiserver.provider import OAIIDProvider
-
 from ..providers.recid import RecordUUIDProvider
+from .opendata import cernopendata_generic_minter
 
 
 def cernopendata_recid_minter(record_uuid, data):
     """Mint deposit's PID."""
-    if "id" in data:
-        recid = data["id"]
```
dianarand, line 11: The new functionality only accepts `recid` field, so are we sure that this `id` field is not used in this way anywhere?

cernopendata/modules/records/minters/opendata.py
```diff
@@ -0,0 +1,35 @@
...
+        data["pids"] = {"oai": {"id": f"oai:cernopendata.cern:{data[id_field]}"}}
+        OAIIDProvider.create(
+            object_type="rec",
+            object_uuid=record_uuid,
+            pid_value=f"oai:cernopendata.cern:{data[id_field]}-v{data['_versions']['index']}",
+        )
+
+    # The first version also registers the concept
+    if data["_versions"]["index"] == 1:
+        provider = provider.create(
```
dianarand, line 23: This might not work as intended, because on line 7 the `provider` variable gets reassigned.

cernopendata/modules/records/minters/opendata.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def cernopendata_generic_minter(
+    record_uuid, data, pid_type, id_field, provider, oai=False
+):
+    provider = provider.create(
+        object_type="rec",
+        pid_type=pid_type,
+        object_uuid=record_uuid,
+        pid_value=str(f"{data[id_field]}-v{data['_versions']['index']}"),
```
dianarand, line 11: `str()` is redundant here

```suggestion
        pid_value=f"{data[id_field]}-v{data['_versions']['index']}",
```

dianarand (changes requested):
There are some linting issues that need to be fixed and it would be nice to add some tests as well.

cernopendata/modules/records/api.py
psaiz: Thanks, removed

cernopendata/config.py
psaiz: Thanks, done

cernopendata/modules/records/minters/opendata.py
psaiz: Thanks, removed

cernopendata/modules/records/minters/opendata.py
psaiz: Thanks, cleaned up

cernopendata/modules/records/minters/recid.py
psaiz: recid should be defined. If not, it should trigger an error

cernopendata/modules/records/api.py
psaiz: Renamed, thanks

cernopendata/modules/fixtures/cli.py
psaiz: The session commit will be in the line 449, isn't it? That should put the new version as well (or none of them, if there was an error with the new version)

PR #328, opened 2026-05-26
feat(releases): add button for generating DOIs for a staged release

### LINKED ISSUE #326: Releases: Automated DOI creation for records
There already is a [checkbox](https://github.com/cernopendata/cernopendata-portal/blob/988f2a42bdc5a2941848080757cd0cdd675465fb/cernopendata/modules/theme/assets/semantic-ui/js/releases/records/RecordsTable.js#L55) for enabling automated DOI creation for records, however it does not function yet. This could be replaced with a button, which would be available to be pressed before publishing the release. It would then generate the DOIs.

Implement the needed endpoint for generating DOIs for the records in a release ([existing todo](https://github.com/cernopendata/cernopendata-portal/blob/988f2a42bdc5a2941848080757cd0cdd675465fb/cernopendata/modules/releases/views.py#L238)).

If the DOIs are successfully generated, they should be added to the JSON of the records. If any errors happen, they should be presented to the user before they move on to publishing the release.

See [here](https://github.com/cernopendata/cernopendata-portal/blob/988f2a42bdc5a2941848080757cd0cdd675465fb/cernopendata/modules/datacite/cli.py#L79) the existing functionality for generating DOIs via CLI.

When publishing the records, the DOIs should get registered. If any failure happens during this, the publishing should continue and the error messages printed.

dianarand:
Resolves #326

### DIFF
--- a/cernopendata/modules/datacite/utils.py
+++ b/cernopendata/modules/datacite/utils.py
@@ -23,11 +23,40 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 """Utilities for datacite module."""
 
+import os
 import random
 import string
 
+from datacite import schema43
+from flask import current_app
 from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_pidstore.models import PersistentIdentifier
+
+from .serializers import DataCiteSerializer
+
+
+def validate_record(record):
+    """Serialize a record to schema43 and validate it."""
+    doc = DataCiteSerializer().dump(record)
+    schema43.validate(doc)
+    return schema43.tostring(doc)
+
+
+def register_record_doi(record_data):
+    """Register a record's DOI with DataCite."""
+    from .providers import DataCiteProviderWrapper
+
+    doi = record_data["doi"]
+    experiment = record_data.get("experiment")
+    try:
+        provider = DataCiteProviderWrapper.get(pid_value=doi, pid_type="doi")
+    except PIDDoesNotExistError:
+        provider = DataCiteProviderWrapper.create(pid_value=doi, experiment=experiment)
+    doc = validate_record(record_data)
+    landing_page = os.path.join(
+        current_app.config["PIDSTORE_LANDING_BASE_URL"], str(record_data["recid"])
+    )
+    provider.register(url=landing_page, doc=doc)
 
 
 def generate_doi(prefix, experiment=None):
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -39,6 +39,11 @@ from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import selectinload
 from sqlalchemy.orm.attributes import flag_modified
 
+from cernopendata.modules.datacite.utils import (
+    generate_doi as mint_doi,
+    register_record_doi,
+    validate_record as validate_datacite_record,
+)
 from cernopendata.modules.fixtures.cli import (
     create_doc,
     create_record,
@@ -483,11 +488,18 @@ class Release:
             raise RuntimeError("Release is not STAGED")
 
         indexer = RecordIndexer()
+        errors = []
         for record_data in self._metadata.records:
             pid_object = PersistentIdentifier.get("recid", record_data["recid"])
             record = update_record(pid_object, record_data, True)
             record.commit()
             indexer.index(record)
+            if record_data.get("doi"):
+                try:
+                    register_record_doi(record_data)
+                except Exception as error:
+                    current_app.logger.exception("DataCite registration failed")
+                    errors.append({"recid": record_data["recid"], "error": str(error)})
 
         for i, doc_data in enumerate(self._metadata.documents or []):
             slug = doc_data.get("slug")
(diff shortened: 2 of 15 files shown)


cernopendata/modules/datacite/utils.py
```diff
@@ -23,12 +23,41 @@
...
+def validate_record(record):
+    """Serialize a record to schema43 and validate it."""
+    doc = DataCiteSerializer().dump(record)
+    schema43.validate(doc)
+    return schema43.tostring(doc)
+
+
+def register_record_doi(record_data):
+    """Register a record's DOI with DataCite."""
+    from .providers import DataCiteProviderWrapper
```
psaiz, line 47: why here and not at the top?

cernopendata/modules/releases/api.py
```diff
@@ -39,6 +39,11 @@
 from sqlalchemy.orm import selectinload
 from sqlalchemy.orm.attributes import flag_modified
 
+from cernopendata.modules.datacite.utils import (
+    generate_doi as mint_doi,
```
psaiz, line 43: why renaming it?

cernopendata/modules/releases/api.py
```diff
@@ -592,20 +605,24 @@ def bulk_update(self, updates, current_user):
...
-        if self.valid_doi:
-            raise RuntimeError("RECIDs already generated")
-
-        for record in self.records:
-            if "doi" not in record:
-                record["doi"] = f"FAKE DOI FOR  {self.experiment}"
-                # This is to tell alchemy that the field has been modified
-                flag_modified(self, "records")
-        self.validate()
+        for record in self._metadata.records:
```
psaiz: This is going to generate a DOI for all the records, isn't it? What if only some of them need a DOI?

cernopendata/modules/releases/api.py
```diff
@@ -592,20 +605,24 @@ def bulk_update(self, updates, current_user):
...
+                record["doi"] = mint_doi(prefix, experiment)
+
+        flag_modified(self._metadata, "records")
+
+        errors = []
+        for record in self._metadata.records:
+            try:
+                validate_datacite_record(record)
+            except Exception as error:
+                errors.append({"recid": record.get("recid"), "error": str(error)})
```
psaiz, line 623: There are two loops here, the first one that would generate the doi for some records, and the second one that validates all records.  What about making a single loop, that validates first the records that do not have a DOI, and if the validation works, create the DOI?  If the DOI is required for the validation, then, the other way around, doing the validate_datacite_record just after the doi have been genereated

cernopendata/modules/releases/views.py
```diff
@@ -262,26 +263,23 @@ def release_json(experiment, release_id):
...
-#    flash(f"Recid created for records in release {release.id}.", "success")
-#
-#    return redirect(f"/releases/{experiment}/{release_id}")
+@blueprint.route(
+    "/releases/<experiment>/<int:release_id>/generate_doi",
+    methods=["POST"],
+)
+@login_required
+def generate_doi(experiment, release_id):
+    """Mint DOIs for all records in a staged release and validate them."""
```
psaiz: This route should receive a list of records that do not need the doi (or the other way around, with the list of records that do need the doi)

psaiz (changes requested):

cernopendata/modules/datacite/utils.py
dianarand: Having it at the top caused a circular import, because `providers.py` imports from this module

cernopendata/modules/releases/api.py
dianarand: There is a method with the same name `generate_doi` in this file already

cernopendata/modules/releases/api.py
dianarand: On the next line we check whether the record already has a DOI. The DOIs get generated for only the records that don't already have one.

cernopendata/modules/releases/views.py
dianarand: Thanks for the recommendation, I changed it now so that this route receives a list of records needing a DOI.

cernopendata/modules/releases/api.py
dianarand: Based on the changes mentioned in the other comment, now this method gets a list of recids for which records the DOIs have to be generated.

cernopendata/modules/releases/api.py
dianarand: I changed it into one loop. I tested it out and having the validation first caused some errors, so I kept it as generating the DOI first, and then validating, but within a single loop.

cernopendata/modules/releases/api.py
```diff
@@ -591,21 +604,25 @@ def bulk_update(self, updates, current_user):
...
-            if "doi" not in record:
-                record["doi"] = f"FAKE DOI FOR  {self.experiment}"
-                # This is to tell alchemy that the field has been modified
-                flag_modified(self, "records")
-        self.validate()
+        errors = []
+        for record in self._metadata.records:
+            if record.get("recid") not in recids:
+                continue
+            record["doi"] = mint_doi(prefix, experiment)
```
psaiz, line 618: What about checking if the record has a doi before minting one?

dianarand: I added a check above here, so the minting happens only if the record does not have a doi yet.

psaiz (approved):

PR #336, opened 2026-06-15
Delete only releases in DRAFT or READY

psaiz:


cernopendata/modules/releases/api.py
```diff
@@ -334,6 +334,12 @@ def _delete_release_images(self):
 
     def delete(self):
         """Delete a release."""
+        if not self.is_status(ReleaseStatus.DRAFT) and not self.is_status(
+            ReleaseStatus.READY
+        ):
+            raise RuntimeError(
```
dianarand, line 340: Now that there can be an error here, it might be nice to handle in a try/except block where this delete method is called (in `cernopendata/modules/releases/views.py` line 295).

dianarand (changes requested):

cernopendata/modules/releases/api.py
psaiz: Thanks! Added

dianarand (approved):

PR #342, opened 2026-06-17
feat(releases): add DOI validation

### LINKED ISSUE #339: Releases: DOI validation
The DOI validation during the release process should check that the suffix is unique, and that the prefix matches the DATACITE_DOI_PREFIX.

In case the DOI_PREFIX does not match, I'm wondering if the test should fail, or if we should introduce the possibility of a test returning a warning.

dianarand:
Resolves #339 

In case the prefix does not match, the validation fails, but there is a possibility to automatically fix it (the prefix gets replaced, the suffix is kept).

### DIFF
--- a/cernopendata/modules/releases/validations/doi.py
+++ b/cernopendata/modules/releases/validations/doi.py
@@ -1,0 +1,99 @@
+"""Validation that record DOIs use the correct prefix and have unique suffixes."""
+
+from flask import current_app
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from ...datacite.utils import generate_doi
+from .base import Validation
+
+
+class ValidDoi(Validation):
+    """Check that record DOIs use the correct prefix and have unique suffixes."""
+
+    name = "Valid DOI"
+    error_message = "DOIs must use the correct prefix and have a unique suffix."
+    applies_to = {"records"}
+
+    def _validate_prefix(self):
+        """Return True only on the production instance, where DOI prefixes are validated."""
+        return current_app.config["INSTANCE_NAME"] == "opendata"
+
+    def validate(self, release):
+        """Check the prefix and suffix uniqueness of every record that has a DOI."""
+        prefix = str(current_app.config.get("PIDSTORE_DATACITE_DOI_PREFIX"))
+        validate_prefix = self._validate_prefix()
+        errors = []
+        used_suffixes = {}
+        for i, record in enumerate(release.records or []):
+            doi = record.get("doi")
+            if not doi:
+                continue
+            doi_prefix, separator, suffix = doi.partition("/")
+            if not separator or not suffix:
+                errors.append(f"Entry {i + 1}: Malformed DOI '{doi}'")
+                continue
+            if validate_prefix and doi_prefix != prefix:
+                errors.append(
+                    f"Entry {i + 1}: DOI prefix '{doi_prefix}' does not match "
+                    f"the prefix '{prefix}'"
+                )
+            if suffix in used_suffixes:
+                errors.append(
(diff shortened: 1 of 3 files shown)


cernopendata/modules/releases/validations/doi.py
```diff
@@ -0,0 +1,82 @@
...
+        for record in release.records or []:
+            doi = record.get("doi")
+            if not doi:
+                continue
+            doi_prefix, separator, suffix = doi.partition("/")
+            if separator and suffix and doi_prefix != prefix:
+                record["doi"] = f"{prefix}/{suffix}"
+        return self.validate(release)
+
+    def fixable(self, release):
```
psaiz: I'm worried about this, and I'm not sure how to do it.

With this validation/fix, all the existing records will fail when we put them in our qa/dev instance because they have a different DOI_PREFIX.  And the fix will overwrite the DOI, making it to point to a DOI that is only for development. 

I see two options, and I don't really like any of the two, so I'd like to hear more alternatives:
* Hardcode the DOI_PREFIX  of production in the validation. Then, if it  validation confirms that PIDSTORE_DATACITE_DOI_PREFIX is indeed production, then it should validate and fix. If the DOI_PREFIX in the validation is different than the one from the configuration, then it knows that it is in development, and it should not overwrite dois
* Make the validation optional, and count on the curator to enable it if needed

psaiz (changes requested):

cernopendata/modules/releases/validations/doi.py
dianarand: How about we have a config value telling us whether it is dev/qa or prod instance, and based on that, only on prod we validate and fix the prefix? On dev/qa we could only check whether the suffix is valid.

dianarand: I changed it so that only in the prod instance the prefix gets validated and fixed.

cernopendata/modules/releases/validations/doi.py
```diff
@@ -0,0 +1,91 @@
...
```
psaiz: What about putting fixable to True? 

If we do need the function, fix and fixable are almost identical. What about doing an internal function, _check_doi(self, fix=False) 
then, fixable would be something like retun self._check_doi(False), and fix return self._check_doi(True)

psaiz (changes requested):

cernopendata/modules/releases/validations/doi.py
dianarand: We need it, because otherwise the automatic fix button will be shown for errors that are not actually fixable. But I created the internal function to remove the code repetition as suggested, thanks.

cernopendata/modules/releases/validations/doi.py
```diff
@@ -0,0 +1,90 @@
...
+            if validate_prefix and doi_prefix != prefix:
+                errors.append(
+                    f"Entry {i + 1}: DOI prefix '{doi_prefix}' does not match "
+                    f"the prefix '{prefix}'"
+                )
+            if suffix in used_suffixes:
+                errors.append(
+                    f"Entry {i + 1}: Duplicate DOI suffix '{suffix}' "
+                    f"(also used by entry {used_suffixes[suffix] + 1})"
+                )
```
psaiz, line 44: I'm thinking about this one (and the one of line 48). If we detect that the DOI is invalid, would it make sense to automatically create a new one?
That would simplify the `fixable`, because then all issues are fixable.

dianarand: Good point, thanks. I changed it so that an invalid or a duplicate DOI can be automatically fixed by generating a new DOI. This allowed to remove the `fixable`.

psaiz (approved):

PR #351, opened 2026-06-23
feat(releases): enable rucio validation for atlas experiment

dianarand:


cernopendata/modules/releases/validations/rucio.py
```diff
@@ -11,19 +12,46 @@ class RucioDatasets(Validation):
...
+            errors.append(
+                f"Missing X509 proxy for {experiment} (expected {user_proxy})"
+            )
+        if errors:
+            return errors
+        os.environ["RUCIO_HOME"] = rucio_home
+        os.environ["X509_USER_PROXY"] = user_proxy
+        return []
+
+    def fixable(self, release=None):
```
psaiz, line 46: What about leaving the default fixable to True? I mean, the validation will not appear if the experiment is not in the list

dianarand: In that case, if there would be an error for either missing rucio configuration or missing proxy, then there would be the label saying "auto-fixable" along with the button for automatically fixing, which would not actually do anything. I refactored a little bit though.

psaiz: Very good point, thanks!

psaiz (approved):

PR #353, opened 2026-06-24
feat(releases): ensure backwards compability for new format of recids

### LINKED ISSUE #340: Releases: Recid generation
The recid generation should prefix the name of the experiment, and have a counter per experiment. The existing entries should keep the old and the new format (for instance, https://opendata.cern/record/80000 should be similar to https://opendata.cern/record/atlas-80000).

The counter should start with the highest number currently used by each experiment.

Create a script that populates the pidstore_pid table. For every existing record it should add a new entry where the pid_value is `<experiment>-<pid_value>`. Only for the `pid_type='recid'`. As a result, links with both the old and the new format of recids should point to the record.

dianarand:
Resolves #340

### DIFF
--- a/cernopendata/modules/releases/validations/recid.py
+++ b/cernopendata/modules/releases/validations/recid.py
@@ -1,6 +1,10 @@
 """Validation process."""
 
+import re
+
 from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
+from sqlalchemy import Integer, cast
 from sqlalchemy.sql import func
 
 from ..models import ReleaseMetadata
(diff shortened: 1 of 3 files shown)


cernopendata/modules/releases/validations/recid.py
```diff
@@ -17,22 +21,63 @@ class ValidRecid(PIDValidation):
...
+                        func.split_part(PersistentIdentifier.pid_value, "-", 2), Integer
+                    )
+                )
+            )
+            .filter(
+                PersistentIdentifier.pid_type == "recid",
+                PersistentIdentifier.pid_value.like(f"{release.experiment}-%"),
+            )
+            .scalar()
+        ) or 0
```
psaiz, line 63: Instead of doing
a = max( r )
b = max (q)
return max(a,b) +1,

what about 

a = max(r,q)
return a + 1

cernopendata/modules/releases/validations/recid.py
```diff
@@ -17,22 +21,63 @@ class ValidRecid(PIDValidation):
...
 
     def fix(self, release):
         """Assign RECIDs to all records in the release."""
         counter = self.next_recid_start(release)
         duplicates = self._duplicate_pids(release)
+        pattern = self._recid_pattern(release)
 
         for record in release.records:
-            if "recid" not in record or record["recid"] in duplicates:
+            recid = record.get("recid")
```
psaiz, line 74: wouldn't this one fail if recid is not defined?

scripts/migrate_recid_aliases.py
```diff
@@ -0,0 +1,84 @@
...
+
+from invenio_db import db
+from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from invenio_records.models import RecordMetadata
+from sqlalchemy.orm import Session
+
+BATCH_SIZE = 1000
+
+print("Starting script...")
```
psaiz, line 13: I'm happy that we have a script to do this migration. Just to be sure: once we have he pid defined as &lt;experiment>-&lt;number> and &lt;number>, could you please confirm that the following links work in your local instance:

localhost/record/&lt;experiment>-&lt;number>
localhost/record/&lt;number>
localhost/api/records/&lt;experiment>-&lt;number>
localhost/api/records/&lt;number>

psaiz (changes requested):

scripts/migrate_recid_aliases.py
dianarand: Yes, I tested this on my local instance, after running the script, for example these work and point to the same result:
- `http://localhost/record/84` and `http://localhost/record/cms-84`
- `http://localhost/api/records/84` and `http://localhost/api/records/cms-84`

cernopendata/modules/releases/validations/recid.py
dianarand: No, `record.get("recid")` returns `None` by default when there is no `"recid"` in `records`.

cernopendata/modules/releases/validations/recid.py
dianarand: It could be done yes, but I don't think it would make it more efficient: even though it would be just one query, it would be a `UNION` of two different tables. I also think it might be less readable and more complex, without any improvement in efficiency.

psaiz (approved):

PR #355, opened 2026-06-25
fix(releases): Add margin to the button.  Display multiple date_created

psaiz:


cernopendata/modules/releases/validations/recid.py
```diff
@@ -30,9 +30,8 @@ def fix(self, release):
         """Assign RECIDs to all records in the release."""
         counter = self.next_recid_start(release)
         duplicates = self._duplicate_pids(release)
-
         for record in release.records:
-            if "recid" not in record or record["recid"] in duplicates:
+            if not record.get("recid", None) or record["recid"] in duplicates:
```
dianarand: `get()` returns `None` by default
```suggestion
            if not record.get("recid") or record["recid"] in duplicates:
```

dianarand: Also, I think in #353 this already gets fixed.

psaiz: Thanks! These changes have been removed

dianarand (approved):

Issue #356, opened 2026-06-25
Releases: Asynchronous publishing


dianarand:
Publishing a release currently runs synchronously in the request. For a large release this could exceed the proxy timeout and fail, similarly to staging as described in #350.

To make publishing robust and consistent with staging, it should run in the background as a Celery task, similarly to how staging was implemented in #350.

PR #357, opened 2026-06-25
fix(releases): move staging a release into a celery task

### LINKED ISSUE #350: Releases: Asynchronous staging
Currently when staging an release, the user might get a 504 Gateway Timeout error after 30 seconds of loading (tested on opendata-dev). The staging works regardless - after the user reopens the portal and navigates back to the release, it's shown to have successfully been staged.

Instead of having a long loading process and danger of a timeout error, the staging process should run in the background as a celery task.

dianarand:
Resolves #350

### DIFF
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -488,6 +488,15 @@ class Release:
         db.session.add(self._metadata)
         db.session.commit()
 
+    def mark_staging_failed(self, message, current_user):
+        """Record a staging failure and revert the release to READY."""
+        self._metadata.errors = [f"Staging failed: {message}"]
+        self._metadata.num_errors = 1
+        flag_modified(self._metadata, "errors")
+        self.change_status(ReleaseStatus.READY, current_user)
+        db.session.add(self._metadata)
+        db.session.commit()
+
     def publish(self, current_user):
         """Publish a release."""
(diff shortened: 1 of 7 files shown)


cernopendata/modules/releases/api.py
```diff
@@ -488,6 +488,15 @@ def stage(self, current_user):
         db.session.add(self._metadata)
         db.session.commit()
 
+    def mark_staging_failed(self, message, current_user):
+        """Record a staging failure and revert the release to READY."""
+        self._metadata.errors = [f"Staging failed: {message}"]
+        self._metadata.num_errors = 1
+        flag_modified(self._metadata, "errors")
+        self.change_status(ReleaseStatus.READY, current_user)
```
psaiz, line 496: Would it make sense to create a new status for this, like STAGED_FAILED?

psaiz:
It would be nice to do it also for the publish. I'm not sure if in this PR, or in a different one

psaiz (approved):
Couple of minor comments, feel free to ignore them, or address them in a different PR

cernopendata/modules/releases/api.py
dianarand: For now, I don't think we need a separate status. Maybe it's something we can consider in the future, if the need arises.

dianarand:
Similar improvement for publishing will be implemented later (issue #356)

PR #362, opened 2026-06-30
feat(datacite): Add more fields to the datacite exporter: keywords, u…

### LINKED ISSUE #244: Enhance the oai-pmh interface to show more information
The information available in the [resource hub](https://open-science-cloud.ec.europa.eu/resources/data?linkedDatasourceLocalID=10%7Cfairsharing_::6f8caa0e6413027cb7a12f945151cb8d&linkedDatasourceName=CERN%20Open%20Data%20Portal) misses a lot of details: no description, keywords and things like that. 

All that information should be send in the oai-pmh interface so that openaire can harvest it.

psaiz:
…sage, methodology and language

### DIFF
--- a/cernopendata/modules/datacite/serializers.py
+++ b/cernopendata/modules/datacite/serializers.py
@@ -28,6 +28,8 @@ from flask import request
 from lxml import etree
 from marshmallow import Schema, fields
 
+from bs4 import BeautifulSoup
+
 
 def datacite_etree(pid, record):
     """Datacite XML format for OAI-PMH.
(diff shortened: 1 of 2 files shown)


cernopendata/modules/datacite/serializers.py
```diff
@@ -28,6 +28,8 @@
 from lxml import etree
 from marshmallow import Schema, fields
 
+from bs4 import BeautifulSoup
```
dianarand, line 31: In `setup.py` `beautifulsoup4` is only in `tests_require `. Maybe it should be moved to `install_requires`?

cernopendata/modules/datacite/serializers.py
```diff
@@ -57,17 +59,52 @@ class DataCiteSerializer(Schema):
...
+            for v in values:
+                if v:  # avoid empty strings / None
+                    subjects.append({"subject": v})
+        return subjects
 
     def get_description(self, obj):
         """Get the description of the object."""
         descriptions = []
-        if "abstract" in obj and "description" in obj["abstract"]:
+        abstract = obj.get("abstract", {}).get("description")
```
dianarand: I'm not sure if that could be the case, but if the `"abstract"` field exists, but has `None` for value, then this will probably raise an error. Same for line 98 with `"methodology"` and 103 with `"usage"`.

cernopendata/modules/datacite/serializers.py
```diff
@@ -57,17 +59,52 @@ class DataCiteSerializer(Schema):
...
             )
+        desc = obj.get("methodology", {}).get("description")
+        if desc:
+            desc = BeautifulSoup(desc, "html.parser").get_text()
+            descriptions.append({"descriptionType": "Methods", "description": desc})
+
+        usage = obj.get("usage", {}).get("description")
+        if usage:
+            usage = BeautifulSoup(usage, "html.parser").get_text()
+            descriptions.append({"descriptionType": "Others", "description": usage})
```
dianarand: Should it be `"Other"` instead of `"Others"`?

dianarand (changes requested):

cernopendata/modules/datacite/serializers.py
psaiz: Good catch. Changed

cernopendata/modules/datacite/serializers.py
psaiz: Yep, correct again. Thanks!

cernopendata/modules/datacite/serializers.py
psaiz: And 3 for 3. Thanks for the review

cernopendata/modules/datacite/serializers.py
```diff
@@ -57,17 +59,52 @@ class DataCiteSerializer(Schema):
...
                 }
             )
+        if obj.get("methodology") and obj.get("methodology").get("description"):
+            desc = BeautifulSoup(
+                obj.get("methodology").get("description"), "html.parser"
+            ).get_text()
+            descriptions.append({"descriptionType": "Methods", "description": desc})
+
+        usage = obj.get("usage", {}).get("description")
+        if usage:
```
dianarand: Similar change as on the lines 90 and 97.
```suggestion
        if obj.get("usage") and obj.get("usage").get("description"):
```

cernopendata/modules/datacite/serializers.py
```diff
@@ -57,17 +59,52 @@ class DataCiteSerializer(Schema):
...
+                    "description": obj.get("abstract").get("description"),
                 }
             )
+        if obj.get("methodology") and obj.get("methodology").get("description"):
+            desc = BeautifulSoup(
+                obj.get("methodology").get("description"), "html.parser"
+            ).get_text()
+            descriptions.append({"descriptionType": "Methods", "description": desc})
+
+        usage = obj.get("usage", {}).get("description")
```
dianarand:
```suggestion
```

dianarand (approved):

PR #367, opened 2026-07-06
fix(releases): move publishing a release into a celery task

### LINKED ISSUE #356: Releases: Asynchronous publishing

dianarand:
Resolves #356

### DIFF
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -39,9 +39,9 @@ from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import selectinload
 from sqlalchemy.orm.attributes import flag_modified
 
+from cernopendata.modules.datacite.utils import generate_doi as mint_doi
+from cernopendata.modules.datacite.utils import register_record_doi
 from cernopendata.modules.datacite.utils import (
-    generate_doi as mint_doi,
-    register_record_doi,
     validate_record as validate_datacite_record,
 )
 from cernopendata.modules.fixtures.cli import (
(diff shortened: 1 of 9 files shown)


cernopendata/modules/releases/api.py
```diff
@@ -488,19 +488,27 @@ def stage(self, current_user):
         db.session.add(self._metadata)
         db.session.commit()
 
-    def mark_staging_failed(self, message, current_user):
-        """Record a staging failure and revert the release to READY."""
-        self._metadata.errors = [f"Staging failed: {message}"]
+    def _mark_failed(self, action, message, revert_status, current_user):
```
psaiz, line 491: cosmetic: (feel free to ignore). Instead of _mark_failed, what about _rollback? This is used as well when the user clicks on going back for modifications, isn't it?

cernopendata/modules/releases/api.py
```diff
@@ -528,6 +536,12 @@ def publish(self, current_user):
             doc.commit()
             indexer.index(doc)
 
+        self._metadata.errors = [
+            f"DataCite registration failed for recid {error['recid']}: {error['error']}"
```
psaiz: i don't like having this `DataCite` hardcoded here. If, at some point, any other part of this returns errors, it will not be consistent. What about modifying already the error in 525, so that the message has the `DataCite` prefix?

cernopendata/modules/releases/api.py
```diff
@@ -528,6 +536,12 @@ def publish(self, current_user):
...
+            for error in errors
+        ]
+        self._metadata.num_errors = len(errors)
+        flag_modified(self._metadata, "errors")
```
psaiz, line 544: We will get then `Published` releases with errors. We will need a way of checking if those errors have been solved (not as a part of this PR).  Would you mind creating a ticket about that? Would it make sense to have some validations run on Published releases ?

psaiz (changes requested):

cernopendata/modules/releases/api.py
dianarand: No, this is only used for a failed action (staging or publishing) to record an error message and to revert to the previous status. While rollback also reverts to a previous status, other than that, these methods are different. Because of that, I wouldn't want to reuse the word "rollback" here.

cernopendata/modules/releases/api.py
dianarand: Good point, thanks. I modified it as suggested.

cernopendata/modules/releases/api.py
dianarand: Thanks, I created issue #368 for this

psaiz (approved):

PR #370, opened 2026-07-08
perf(releases): optimise migration script for recids

dianarand:


scripts/migrate_recid_aliases.py
```diff
@@ -12,14 +11,28 @@
...
+    )
+    .outerjoin(RecordMetadata, PersistentIdentifier.object_uuid == RecordMetadata.id)
+    .filter(PersistentIdentifier.pid_type == "recid")
+    .filter(PersistentIdentifier.pid_value.op("~")(r"^[0-9]+$"))
+    .order_by(PersistentIdentifier.id)
 )
 
+existing_aliases = {
+    value
+    for (value,) in db.session.query(PersistentIdentifier.pid_value)
```
psaiz, line 28: This query might be huge, isn't it? 

What about putting it as a filter in the other query, something like 
```
.filter(
        ~exists().where(
            (PI2.pid_type == PersistentIdentifier.pid_type)
            & (
                PI2.pid_value.op("~")(
                    func.concat(r"^.+-", PersistentIdentifier.pid_value, "$")
                )
            )
        )
    )
```

dianarand: I think the existing query should be fine, since it only fetches the recid values and nothing else, and its results are stored as a set, which should be good for fast lookups. I think the recommended filter might not improve the performance.

psaiz (approved):

PR #371, opened 2026-07-08
Various fixes

dianarand:
- fix(releases): ensure no recid overlap
- fix(releases): include release metadata in download and creation
- schema: add filename
- fix(releases): ensure export json feature works for creating release

### DIFF
--- a/cernopendata/modules/releases/views.py
+++ b/cernopendata/modules/releases/views.py
@@ -198,7 +198,7 @@ def release_upload(experiment):
             payload = json.load(file)
         except Exception as e:
             return jsonify({"error": "Invalid JSON", "details": str(e)}), 400
-        release_name = file.filename.rsplit("/", 1)[-1]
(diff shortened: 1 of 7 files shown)


cernopendata/modules/releases/views.py
```diff
@@ -254,20 +281,31 @@ def update_release(experiment, release_id):
...
     """Get the release in json."""
     release = _get_release(experiment, release_id)
+    metadata = release._metadata
     body = {
-        "records": release._metadata.records,
-        "documents": release._metadata.documents or [],
+        "name": metadata.name,
+        "description": metadata.description,
+        "discussion_url": metadata.discussion_url,
+        "records": metadata.records,
```
psaiz: why documents and records are treated differently?

dianarand: It seems like there was no particular reason for the difference, I made it consistent now.

psaiz (approved):

PR #373, opened 2026-07-15
fix(releases): ensure file checksums have a prefix

### LINKED ISSUE #372: Releases: Ensure checksums have prefix
Files added to a release through the file metadata validation or through the directory expansion get their checksum as a bare hex value. Instead it should be with the `adler32` prefix, like for example in the rucio validation.

This causes an error when later the release is published and an user attempts to download any of the files of the affected record
```
{"status": 500, "message": null, "errors": "Could not send file: not enough values to unpack (expected 2, got 1)"}
```

Additionally to fixing this bug, the already existing files should have their checksums fixed in the database. These can be found with 
```
SELECT * FROM files_files WHERE checksum NOT LIKE '%:%';
```

dianarand:
Resolves #372

### DIFF
--- a/cernopendata/modules/releases/validations/expand_files.py
+++ b/cernopendata/modules/releases/validations/expand_files.py
@@ -53,13 +53,16 @@ class CheckExpandDirectories(ValidFiles):
 
         for entry in entries:
             full_uri = f"{base_uri}/{entry}"
-            isDir, size, checksum = self._get_entry_details(ctx, f"{http_uri}/{entry}")
+            is_dir, size, checksum = self._get_entry_details(ctx, f"{http_uri}/{entry}")
 
-            # If it’s a directory, recurse
-            if isDir:
+            # If it's a directory, recurse
+            if is_dir:
                 yield from self._walk(ctx, full_uri)
             else:
-                yield {"uri": full_uri, "size": size, "checksum": checksum}
+                file = {"uri": full_uri, "size": size}
+                if checksum:
+                    file["checksum"] = checksum
+                yield file
 
     def validate(self, release):
         """Check if there are any directories as input for a record."""
--- a/cernopendata/modules/releases/validations/files.py
+++ b/cernopendata/modules/releases/validations/files.py
@@ -11,31 +11,44 @@ class ValidFiles(Validation):
     """Check if the files have the proper metadata."""
 
     name = "File metadata"
-    error_message = "Some of the files are missing the size or checksum"
+    error_message = (
+        "Some of the files are missing the size, or have a missing or invalid checksum"
+    )
+
+    @staticmethod
+    def _has_checksum_prefix(checksum):
+        """Return whether the checksum has the ``algorithm:value`` shape with both parts non-empty."""
+        parts = str(checksum).split(":")
+        return len(parts) == 2 and all(parts)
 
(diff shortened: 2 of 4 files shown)


cernopendata/modules/releases/validations/files.py
```diff
@@ -23,19 +25,23 @@ def validate(self, release):
                         errors.append(
                             f"Entry {i + 1}, file {j + 1}: Missing size/checksum"
                         )
+                    elif ":" not in str(file["checksum"]):
```
Soap2G: This is low risk, but: 
I think this would pass even if I added `a:b:c`. Do you think it's worth validating for exactly two non empty parts, or have an allowlist of algorithms somewhere?

cernopendata/modules/releases/validations/expand_files.py
```diff
@@ -53,13 +53,16 @@ def _walk(self, ctx, base_uri):
...
+            if is_dir:
                 yield from self._walk(ctx, full_uri)
             else:
-                yield {"uri": full_uri, "size": size, "checksum": checksum}
+                file = {"uri": full_uri, "size": size}
+                if checksum:
+                    file["checksum"] = checksum
+                yield file
 
     def validate(self, release):
```
Soap2G, line 67: I think that this returns only one issue per file. If a file has multiple issues, they are hidden one behind the other. Worth a chore to make the ifs independent? (maybe for a later commit?)

scripts/fix_bare_checksums.py
```diff
@@ -0,0 +1,104 @@
...
+bucket_ids = {
+    bucket_id
+    for (bucket_id,) in db.session.query(ObjectVersion.bucket_id).filter(
+        ObjectVersion.file_id.in_(bare_file_ids)
+    )
+}
+records_owning_the_files = db.session.query(RecordsBuckets.record_id).filter(
+    RecordsBuckets.bucket_id.in_(bucket_ids)
+)
+records_indexing_the_files = db.session.query(cast(BucketTag.value, UUID)).filter(
```
Soap2G, line 60: Is this cast based on the fact that we are sure that `BucketTag` value is a valid UUID string?
If we are sure, then no issues there.

Soap2G (approved):
Hi @dianarand, lgtm :)
I've left a couple of minor remarks, but if we wanted, we could merge this right away. You let me know.

Also, just as curiosity: what's the coverage policy? I see that there is a decent coverage already; we are missing coverage for `expand_files._walk` I think.

Thanks a lot!

cernopendata/modules/releases/validations/files.py
dianarand: Good point, thanks. I improved this so that only two non empty parts pass the validation.

cernopendata/modules/releases/validations/expand_files.py
dianarand: Also good point, thanks, fixed this.

scripts/fix_bare_checksums.py
dianarand: Yes, verified this based on the records that had files with broken checksums.

Soap2G:
@dianarand it's good to me, but I don't have permission to merge.
If you know someone who does, feel free to ping them :D

PR #380, opened 2026-07-21
Releases table redesign

dianarand:
Resolves #324 
- feat(releases): add table sorting/filtering and redesign the releases page
image

- fix(releases): make discussion link visible in release view
image

PR #384, opened 2026-07-24
perf(releases): move rolling back a release into a celery task

### LINKED ISSUE #365: Releases: Asynchronous rollback
Rolling back a release currently runs synchronously in the request. For a large release this could exceed the proxy timeout and fail, similarly to staging as described in #350.

While testing, an user has reported getting error for 502 Bad Gateway while attempting to rollback, and the action was not completed. For a bigger release, the rollback can take a significant amount of time and display a timeout error to the user, even if eventually the rollback will complete. It can be unclear for the user whether it is working and taking a long time, or it has silently failed already.

To make rolling back robust and consistent with staging and publishing, it should run in the background as a Celery task, similarly to implementations in in #350 and #356

dianarand:
Resolves #365

### DIFF
--- a/cernopendata/modules/releases/api.py
+++ b/cernopendata/modules/releases/api.py
@@ -507,6 +507,10 @@ class Release:
         """Record a publishing failure and revert the release to STAGED."""
         self._mark_failed("Publishing", message, ReleaseStatus.STAGED, current_user)
 
+    def mark_rollback_failed(self, message, current_user):
+        """Record a rollback failure and revert the release to STAGED."""
+        self._mark_failed("Rollback", message, ReleaseStatus.STAGED, current_user)
+
     def publish(self, current_user):
         """Publish a release."""
         if not self.is_status(ReleaseStatus.PUBLISHING):
@@ -551,8 +555,8 @@ class Release:
 
     def rollback(self, current_user):
         """Remove the STAGED entries of a release."""
-        if not self.is_status(ReleaseStatus.STAGED):
-            raise RuntimeError("Release is not STAGED")
+        if not self.is_status(ReleaseStatus.ROLLINGBACK):
+            raise RuntimeError("Release is not ROLLINGBACK")
 
         for record_data in self._metadata.records:
             pid_object = PersistentIdentifier.get("recid", record_data["recid"])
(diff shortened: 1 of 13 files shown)


cernopendata/modules/releases/api.py
```diff
@@ -551,8 +555,8 @@ def publish(self, current_user):
 
     def rollback(self, current_user):
```
Soap2G, line 556: This deletes records/docs in a loop and commits once at the end. 
If it fails midway, `db.session.rollback()` undoes the DB deletes, but Invenio's `delete_record` / `search-index` can still be there, and the release returns to STAGED. 
Re-running rollback then calls `PersistentIdentifier.get("recid", …)` on already-deleted PIDs and can raise `PIDDoesNotExist`. Worth making the deletes idempotent (skip missing PIDs)?

cernopendata/modules/theme/assets/semantic-ui/js/releases/ReleasesTable.js
```diff
@@ -5,6 +5,8 @@ const statusToStep = {
   EDITING: "DRAFT",
   READY: "DRAFT",
   STAGING: "DRAFT",
+  ROLLINGBACK: "STAGED",
+  PUBLISHING: "STAGED",
```
Soap2G, line 18: This stuff should survive #380

Soap2G (commented):
Hi @dianarand, thanks!

Just flagged a couple of items, rest looks great!

cernopendata/modules/theme/assets/semantic-ui/js/releases/ReleasesTable.js
dianarand: Rebased the branch and ensured this survived.

cernopendata/modules/releases/api.py
dianarand: Maybe I'm misunderstanding here, but if it fails midway, then when running it again we do not get `PIDDoesNotExist` because the DB changes got rolled back.

However, when looking at this, I discovered that there was a bug related to calling `delete_record` and `delete_doc_or_glossary` without a logger, so I made some fixes related to that.

Soap2G (approved):

PR #385, opened 2026-07-24
fix(theme): center footer content and pin it to the bottom

### LINKED ISSUE #311: Fix and improve footer placement
On [opendata-dev](https://opendata-dev.cern.ch/), the footer contents are not centered as they were before. See

Image

compared to

Image

This seems to be possibly be an unintended consequence from [this commit](https://github.com/cernopendata/cernopendata-portal/commit/f89fc17b523080bba6334c578e0ebf5301190e0a#diff-cc27c01ea822e878e44a14a45cc9a0bf4d26e1be952a2ff9f51092a84493ba5d).

Additionally, it would look better, if the footer was fixed to the bottom of the page. Currently, if the page content is not long, the footer can be very high up, for example 

Image

Instead of having the empty space below the footer, it might look better to have it above the footer, keeping the footer always at the bottom of the page.

dianarand:
Resolves #311 

image

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/footer.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/footer.scss
@@ -22,31 +22,76 @@
 
 $footer-height: 138px;
 
+body {
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+}
+
+.footer-wrap {
+  flex: 1 0 auto;
+}
+
 .footer {
+  flex-shrink: 0;
+  min-height: $footer-height;
   background-color: white;
   border-top: 2px solid #eee;
 
-  .ui.list:first-child.logos {
-    margin-top: 20px
+  .footer-inner {
+    $logos-min: min(26rem, 32%);
+
+    grid-template-columns:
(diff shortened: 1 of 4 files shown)


cernopendata/modules/theme/assets/semantic-ui/scss/footer.scss
```diff
@@ -22,31 +22,38 @@
 
 $footer-height: 138px;
 
+body {
```
Soap2G, line 25: This is a global change; if it hits every page, it's worth checking that the rest looks ok after this change?

Soap2G (approved):
Hi @dianarand, lgtm thanks!

Just one small question.

cernopendata/modules/theme/assets/semantic-ui/scss/footer.scss
dianarand: Everything looks okay, I looked at various pages to verify.

However, when rechecking I noticed a minor issue on the search results page that I now fixed in `search.scss` and I also improved the changes in `footer.scss` so that it looks better on smaller screen sizes.

Soap2G (approved):

PR #386, opened 2026-07-28
refactor(releases): move release styles into their own stylesheet

dianarand:
Just moving the release styles into a separate stylesheet so that they do not get loaded unless needed. No other changes.


Soap2G (approved):
lgtm, thanks @dianarand

PR #387, opened 2026-07-28
fix(releases): return a 404 page for a non-existent release

### LINKED ISSUE #369: Releases: Proper error page for non-existent release
If going to a page for a non-existent release, for example by using the URL `https://opendata-dev.cern.ch/releases/atlas/<non-existent-release-id>`, then there will be a 502 Bad Gateway error. There should instead be a proper error page.

See if there are any other similar cases that could also have a proper error page.

dianarand:
Resolves #369


Soap2G (approved):
lgtm!

PR #400, opened 2026-08-10
Quick fixes

### LINKED ISSUE #399: Releases: Downloading a release adds unnecessary fields to its records
Downloading a release includes fields that were not in the JSON used to create the release: 
`_availability_details`, `_bucket`, `_file_indices`, `_files`, `dataset`.

Using that file to create a release carries those fields into the new release. This can cause issues when indexing when publishing (for example in the local instance)
```
Publishing failed: RequestError(400, 'strict_dynamic_mapping_exception', 'mapping set to strict,
dynamic introduction of [dataset] within [_doc] is not allowed')
```
These fields could be stripped when downloading the release or uploading a release. The downloaded release should match what was uploaded to create it.

dianarand:
- fix(mappings): map methodology.steps.note as an object
- fix(datacite): fall back to the collaboration when authors is empty
- fix(releases): strip the derived fields from records on import and export (resolves #399)
- fix(fixtures): keep the record files when updating with skip_files
- fix(releases): stop long release names overflowing the releases table

### DIFF
--- a/cernopendata/modules/datacite/serializers.py
+++ b/cernopendata/modules/datacite/serializers.py
@@ -121,7 +121,11 @@ class DataCiteSerializer(Schema):
 
     def get_creator(self, obj):
         """Get creators based on authors or collaboration field."""
-        authors = obj.get("authors", [obj.get("collaboration", None)])
+        authors = obj.get("authors") or []
+        if not authors:
+            collaboration = obj.get("collaboration") or {}
+            if collaboration.get("name"):
+                authors = [collaboration]
         creators = [
             {
                 "name": author["name"],
@@ -138,6 +142,7 @@ class DataCiteSerializer(Schema):
                 ),
             }
             for author in authors
+            if author.get("name")
         ]
         return creators
 
--- a/cernopendata/modules/fixtures/cli.py
+++ b/cernopendata/modules/fixtures/cli.py
@@ -161,6 +161,7 @@ def create_record(data, skip_files, logger=None):
 def update_record(pid, data, skip_files, logger=None):
     """Updates the given record."""
     record = RecordFilesWithIndex.get_record(pid.object_uuid)
+    file_keys = ["files", "_files", "file_indices", "_file_indices"]
     if not skip_files:
         for o in ObjectVersion.get_by_bucket(record.bucket).all():
             o.remove()
@@ -171,10 +172,13 @@ def update_record(pid, data, skip_files, logger=None):
     for k in list(record.keys()):
         if k in ["_bucket", "pids"]:
             continue
-        if skip_files and k in ["files", "_files", "file_indices", "_file_indices"]:
+        if skip_files and k in file_keys:
             continue
         del record[k]
-    record.update(data)
+    if skip_files:
+        record.update({k: v for k, v in data.items() if k not in file_keys})
+    else:
+        record.update(data)
     if not skip_files:
         _handle_record_files(record, data, logger)
         record.update(data)
--- a/cernopendata/modules/releases/views.py
+++ b/cernopendata/modules/releases/views.py
@@ -102,6 +102,23 @@ def _merge_index_files(record):
     record["files"] = files
 
 
+def _strip_derived_fields(record):
+    """Return the record without the fields derived when it was created."""
+    if not isinstance(record, dict):
+        return record
+    derived_fields = (
+        "_availability_details",
+        "_bucket",
+        "_file_indices",
+        "_files",
+        "bucket",
+        "dataset",
+    )
+    return {
+        field: value for field, value in record.items() if field not in derived_fields
+    }
+
+
 def _normalise_record(record):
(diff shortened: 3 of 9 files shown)


cernopendata/modules/datacite/serializers.py
```diff
@@ -121,7 +121,11 @@ def get_identifiers(self, obj):
 
     def get_creator(self, obj):
         """Get creators based on authors or collaboration field."""
-        authors = obj.get("authors", [obj.get("collaboration", None)])
+        authors = obj.get("authors") or []
+        if not authors:
+            collaboration = obj.get("collaboration") or {}
+            if collaboration.get("name"):
+                authors = [collaboration]
```
psaiz, line 128: Is there a way to return an error if we get here, and there is no collaboration?

cernopendata/modules/fixtures/cli.py
```diff
@@ -171,10 +172,13 @@ def update_record(pid, data, skip_files, logger=None):
     for k in list(record.keys()):
         if k in ["_bucket", "pids"]:
             continue
-        if skip_files and k in ["files", "_files", "file_indices", "_file_indices"]:
+        if skip_files and k in file_keys:
             continue
         del record[k]
-    record.update(data)
+    if skip_files:
+        record.update({k: v for k, v in data.items() if k not in file_keys})
```
psaiz, line 179: Was this the responsible for _file being empty in https://opendata.cern/record/atlas-160004 ?

cernopendata/modules/releases/views.py
```diff
@@ -102,15 +102,31 @@ def _merge_index_files(record):
...
+    """Return the record without the fields derived when it was created."""
+    if not isinstance(record, dict):
+        return record
+    derived_fields = (
+        "_availability_details",
+        "_bucket",
+        "_file_indices",
+        "_files",
+        "bucket",
+        "dataset",
```
psaiz, line 115: Are you sure about the two fields that do not start with '_' ?

cernopendata/modules/releases/views.py
```diff
@@ -102,15 +102,31 @@ def _merge_index_files(record):
...
+
+
 def _normalise_record(record):
     """Normalise a single exported record for release import."""
     if not isinstance(record, dict):
         return record
     # In case we are reading from the cernopandata api, where the record is in the 'metadata' field
     if "metadata" in record:
         record = record["metadata"]
-        for field in ("_files", "_bucket", "bucket", "_file_indices"):
```
psaiz, line 112: uhm, I wonder if "bucket" was supposed to be there...

cernopendata/modules/releases/views.py
```diff
@@ -133,8 +149,8 @@ def _split_payload(payload, source_filename):
         isinstance(payload.get("records"), list)
         or isinstance(payload.get("documents"), list)
     ):
-        records = payload.get("records") or []
-        documents = payload.get("documents") or []
+        records = _normalise_payload(payload.get("records") or [])
+        documents = _normalise_payload(payload.get("documents") or [])
```
psaiz: documents should not have files, file_indices, buckets or availabilities. Do they need to be normalized?

cernopendata/modules/fixtures/cli.py
dianarand: I believe so. Hopefully this change should avoid this happening in the future, but atlas-160004 will have to be manually fixed.

cernopendata/modules/releases/views.py
dianarand: The `"bucket"` field has been excluded since first release upload commit, and it is not in the record schema at all, so if we do not strip it, the record will fail schema validation.
The `"dataset"` field however is in the record schema, but does not seem to be actually in use anywhere, not a single already existing record has it, and it causes indexing to fail - I described this in the issue #399

cernopendata/modules/releases/views.py
dianarand: Talked about this in the reply to the previous comment.

cernopendata/modules/releases/views.py
dianarand: You're right, I removed this. Thanks

cernopendata/modules/datacite/serializers.py
dianarand: I added the error to `cernopendata/modules/datacite/utils.py`

psaiz: Not for this PR: what do you think about having a `datacite` validation, that will make sure that the serializer of datacite produces something reasonable? For instance, it should complain if the authors and the collaboration is empty

psaiz (approved):

cernopendata/modules/datacite/serializers.py
dianarand: Good point, thanks. Created an issue for this #402

PR #415, opened 2026-09-11
fix(news): add CORS headers to news endpoint (closes #354)

### LINKED ISSUE #354: CORS headers for news endpoint
The "News & Updates" panel on the Grafana dashboard ([see here](https://monit-grafana.cern.ch/d/da06d76c-24f0-4d23-b51e-da08d36c4ece/welcome?orgId=93)) fails to load with the error:
```
Error loading RSS feed
Make sure that the feed URL is correct and that CORS is configured correctly on the server.
```
Currently, the panel is configured to read the feed via a third-party proxy:
`https://api.allorigins.win/raw?url=https://opendata.cern.ch/api/news.xml`

The reason for the failure seems to be the allorigins proxy. It seems to be used to inject a missing CORS header. Adding that to the the news endpoint response in `cernopendata/modules/api/news.py` should allow using the endpoint directly, without relying on any third-party proxy.

Cid-oe:
Closes #354

### Summary
The Grafana dashboard "News & Updates" panel failed to load the RSS feed directly from `https://opendata.cern.ch/api/news.xml` because the endpoint was missing the `Access-Control-Allow-Origin` header, forcing external callers to rely on third-party CORS proxies like `allorigins.win`.

### Changes
- Attached `response.headers["Access-Control-Allow-Origin"] = "*"` to the Response returned by `get_latest_news()` in `cernopendata/modules/api/news.py`.
- Added a test `test_news_endpoint_cors` to `tests/test_news.py` ensuring the header is present and valid.


Cid-oe:

PR #419, opened 2026-09-12
fix(ui): add loading feedback and error handling to files section

### LINKED ISSUE #406: Files and indexes section gives no feedback while the file list is loading
When opening a record details page, sometimes in the "Files and indexes" section, the file table takes some seconds to load. During this time, there is no spinner, message or anything else to let the user know that the file table is being loaded, so the section can look like the files are just missing.

For example, when opening the page:

Image

Some seconds later:

Image

> Source: [record 35630](https://opendata.cern.ch/record/35630)

It would be nice to have a loading placeholder (for example a spinner or a skeleton table) while fetching this table, so that it is clearly displayed to the user that it is being loaded. In case it fails, there should be an error message.

Cid-oe:
Closes #406

- Displays a loading indicator while files and indexes are being retrieved asynchronously.
- Handles fetch errors gracefully with an informative error message.
- Adds changelog entry to CHANGES.rst.


Cid-oe:

PR #420, opened 2026-09-12
fix(cold): ensure transfers table resolves recid pid type

### LINKED ISSUE #377: Cold storage: transfers table shows OAI identifiers instead of recids
On [opendata-dev](https://opendata-dev.cern.ch/transfer_requests), the record column of the transfer requests table shows values like `oai:cernopendata.cern:75485` instead of the recid `75485`. This breaks the record links and the subscribe action. 

Image

Currently, QA and production instances do not experience this issue.

Cid-oe:
Closes #377

In `TransferRequestSchema.get_recid`, filter `PersistentIdentifier` specifically by `pid_type="recid"` so that other identifier types (such as `oai`) sharing the same `object_uuid` do not return OAI identifier strings instead of the numeric record ID.


Cid-oe:

PR #421, opened 2026-09-12
fix(releases): prevent HTML form editor from creating empty fields

### LINKED ISSUE #409: curator HTML form editor adds empty fields
* Start with a release, like https://opendata-dev.cern.ch/releases/atlas/32
* Click on `Edit` to modify the single record of this release
* Move to the `Raw Editor` tab, and confirm that the field `author` does not exist
* Go back to the `Form` tab, and change the value of any field (like adding a `hello` in the publisher)
* Go back to the `Raw Editor` tab, and now the field `authors' (with an empty list) is there

Those empty fields should not be created

Cid-oe:
Closes #409

- Adds `pruneEmpty` utility in `releases/shared/utils.js` to recursively strip empty strings, empty arrays, and empty objects from the model.
- Applies `pruneEmpty` in `EditRecordModal.js` and `EditDocumentModal.js` so that editing fields in the HTML Form tab does not inject empty lists/objects (such as `authors: []`) into the model and raw editor.


dianarand:
Issue #409 has already been addressed in PR #410

PR #422, opened 2026-09-12
fix(records): improve error message display on failed stage requests

### LINKED ISSUE #405: Getting an error when requesting all files of a partial release
Hi developers!

I (and others) are seeing an error when we use the "Request all files" button for a dataset with partial availability. I haven't checked all of them, but I see the behavior for this dataset. 

https://opendata.cern.ch/record/35630

I can click on the button and fill in the fields in the popup (I confirm, I enter my email to be notified), but then when I click "OK",  I get a new popup with an error message and "Forbidden" in the HTML. 

Image

I'm using Chrome 149.0.7827.102 on Debian Linux and the behavior is the same both when I'm logge into Chrome and in an incognito tab. 

Matt

Cid-oe:
Closes #405

- Cleanly parses error responses in `RequestRecord.js` instead of rendering raw HTML strings or crashing when non-JSON error payloads (such as 403 Forbidden or server errors) are returned.
- Provides specific, user-friendly feedback for HTTP 403 (Access forbidden) and status-based errors.
- Updates CHANGES.rst.


dianarand:
Issue #405 has already been addressed in PR #408

PR #423, opened 2026-09-12
fix(cold): ensure staging requests only complete when online

### LINKED ISSUE #378: Cold storage: Requests can be marked as completed without any files being staged
After deploying a quick fix for the issue described in #376 in which the schema was relaxed to allow `string` values for `methodology.steps[].note`, an existing bug was exposed, causing requests that were previously stuck to incorrectly be marked as completed:

Seven staging requests were marked as completed and notification emails were sent to their subscribers, but no files had been actually transferred and the files were still on tape. For example `cernopendata cold list 75161` reported 1 hot copy out of 22 files. 

Furthermore, an eighth request was only partially completed when it was set to completed, and in the transfer requests table, the number of successful transfers is incorrectly displayed to be a negative number.

Image

> Source: [Transfer requests](https://opendata.cern.ch/transfer_requests)

Ensure that:
- a staging request is only completed when the record availability is actually online
- a success email is not sent out for a request that did not stage the files
- a started request has issued transfers
- number of successful transfers is displayed correctly in the transfers table

Cid-oe:
Ensure staging requests only complete when record availability is actually online and transfers table shows valid counts.

Closes #378


Cid-oe:

PR #424, opened 2026-09-12
fix(fixtures): preserve cold storage information when updating records

### LINKED ISSUE #413: Cold storage: Updating a record removes its cold storage information
Several records had metadata that needed to be updated to match a schema change. The JSON file were updated and reloaded with `cernopendata fixtures records -f <updated_json>`. Many of the files of these records were on tape, and after the reload the portal no longer showed that they had to be requested. The download button appeared enabled, and clicking it returned:

```
{"status": 500, "message": null, "errors": "Could not send file: resource 'root://eospublic.cern.ch//eos/opendata/cms/MonteCarlo2012/Summer12_DR53X/DYJetsToLL_M-10to50_HT-200to400_TuneZ2star_8TeV-madgraph-tauola/AODSIM/PU_S10_START53_V19-v1/30000/0C817C33-7017-E411-B50C-002590747D94.root' not found"}
```

To fix the records, for each record, these commands had to be run:

```console
cernopendata cold archive --register <recid>
cernopendata cold clear-hot <recid> --limit -1 --dry
```

However for a large amount of records, this can take an extensive amount of time.

This seems to happen because when running `cernopendata fixtures records -f <updated_json>`,  every `ObjectVersion` of its bucket is removed. The cold storage information of a file lives only in `ObjectVersionTag` on its object version, so removing the object version deletes them as well. 

Each file is then recreated and a hardcoded `"availability": "online"` is added. This seems to happen even when using `--skip-files`.

Ensure that updating a record with `cernopendata fixtures records` preserves the `uri_cold` and `hot_deleted` tags of its files and file indices. Alternatively, ensure that there is an easier/quicker way of restoring the cold storage information in case it is not feasible to ensure that the tags are preserved.

Cid-oe:
Preserves cold storage tags and availability details when updating records with or without skipping files.

Closes #413


Cid-oe:

PR #425, opened 2026-09-12
fix(search): enable sorting on glossary terms (closes #318)

### LINKED ISSUE #318: Sorting does not work on glossary term
Image

Check the image. Selecting only the glossaries, and then changing the sorting (like Z-A for instance) does not affect the results

Cid-oe:
### Description
When filtering by type Glossary in search results, sorting by "Title A-Z" or "Title Z-A" had no effect because glossary records store the term in the `anchor` field rather than `title`, causing OpenSearch to treat `title` as empty/missing.

### Changes
- Updated `cernopendata/jsonschemas/records/glossary-term-v1.0.0.json` to include an optional `title` property.
- Updated `load_glossary_data` in `cernopendata/modules/fixtures/cli.py` to populate `title` with `anchor` when loading glossary terms into records.
- Updated `CHANGES.rst`.

Closes #318.


Cid-oe:

PR #431, opened 2026-09-12
feat(releases): validate DataCite metadata of records with DOIs (closes #402)

### LINKED ISSUE #402: Releases: Validate the DataCite metadata of records
A user encountered a bug when publishing a release, the DOI failed to register. The publishing itself was successful.

Image

This was caused by the record containing `"authors": []`. A quick fix was created for this, ensuring that there is a fallback or an error (see [here](https://github.com/cernopendata/cernopendata-portal/pull/400/changes/6e2292d45d2f45810eea00806ea8831ca45b123e)).

It would be better to additionally have a validation for releases that would ensure that the serializer for DataCite will return a valid result for a successful DOI registration. This would allow to avoid any problems or errors already before staging the release.

Something to consider: currently it is possible to release records without adding DOIs, so does it make sense to have the validation in that kind of case?

Cid-oe:
### Description
When publishing a release, records with invalid DataCite metadata (such as records with empty `authors: []` and no collaboration) cause DOI registration to fail during release publication.

This PR adds DataCite metadata validation to the `Valid DOI` validation rule (`ValidDoi`). It verifies that all records with DOIs can be serialized and pass DataCite schema43 validation ahead of time, catching any registration issues before staging/releasing.

### Changes
- In `cernopendata/modules/releases/validations/doi.py`, call `validate_record(record)` on records containing a DOI in `ValidDoi.validate`.
- Append informative error messages including record identifiers if DataCite validation fails.
- Update unit tests in `tests/releases/test_validator_doi.py` to cover DataCite metadata validation.
- Update `CHANGES.rst` referencing closes #402.

Closes #402.


Cid-oe:

PR #433, opened 2026-09-12
fix(records): fall back to record title for parent dataset relation (closes cernopendata/opendata.cern.ch#3707)

### LINKED ISSUE #3707: CMS: Fix relationship tag between NANO and MINI for 2016 collision datasets
If ones look at for example [/Tau/Run2016H-UL2016_MiniAODv2-v1/MINIAOD](https://opendata.cern.ch/record/30548) one can see that there is a "ParentDataset:" tag which links to the corresponding NANOAOD dataset [/Tau/Run2016H-UL2016_MiniAODv2_NanoAODv9-v1/NANOAOD](https://opendata.cern.ch/record/30565). There is no corresponding tag link back to MINIAOD. 

In the json this is governed by the `isChildOf` field. Strictly speaking the relationship between NANOAOD and MINIAOD should be a sibling relationship since they are both produced from the AOD(?) so the field in the json should be `isSiblingOf`. Regardless, the NANO is not the parent of MINI. 

We can either remove this altogether since there are already links in the record or use the `isSiblingOf` field properly.

Cid-oe:
Closes cernopendata/opendata.cern.ch#3707

### Description
In record detail templates (`record_detail.html` and `record_detail_Author-Lists.html`), when a relation entry has `type == 'isChildOf'` and a `recid`, fall back to `relation.recid | get_record_title` if `relation.title` is absent.

This resolves the issue where records having relations with a `recid` and `description` but without an explicit `title` property (such as 2016 collision datasets) displayed an empty `Parent Dataset: ` label without a dataset name.


Cid-oe:

Issue #435, opened 2026-09-17
Cold storage: Duplicate staging requests


dianarand:
It is currently possible to have duplicate staging requests, causing confusing information in the transfer requests table. The first request is completed as expected, and therefore the information about issued transfers is shown. However the second request does not issue any transfers, as they were already completed by the first request, causing a confusing row in the table.

<img width="1332" height="263" alt="Image" src="https://github.com/user-attachments/assets/3dcf070e-2b79-430d-bdfb-e0d48aa20d67" />

> Source: [Transfer requests for record 35630](https://opendata.cern.ch/transfer_requests?record_id=35630)

Furthermore, the failure count is not working as intended, as also outlined in issue #378. This is caused by the failure count being incorrectly per record instead of per request, therefore it causes a negative number to be shown in the case of duplicate requests, but also any repeated requests for the same record.

Ensure that:
- A staging request is not created when the same record already has a request for the same action and the same scope
- The user requesting a record that is already being staged gets notified accordingly, and is shown the option to subscribe to the existing request

PR #436, opened 2026-09-17
fix(cold): only count failures of current transfer

dianarand:
Assisted-by: Claude Opus 5 (medium effort)

Bug mentioned in #378 and #435


cernopendata/cold_storage/api.py
```diff
@@ -164,11 +164,13 @@ def get_active_transfers_threshold(action):
             return current_app.config["COLD_ACTIVE_ARCHIVING_TRANSFERS_THRESHOLD"]
 
     @staticmethod
-    def get_failed_transfers_count(record_id):
+    def get_failed_transfers_count(request):
         """Get number of failed transfers."""
         return TransferMetadata.query.filter(
-            TransferMetadata.record_uuid == str(record_id),
+            TransferMetadata.record_uuid == str(request.record_id),
```
psaiz: let's go through this scenario:
* A user submits a request to transfer 100 files of record R
* Then, there is another transfer to transfer another 100 files of the same record R

Given that the methos is counting based only on the record id, and not the transfer, is there going to be double counting?

dianarand: So far the transfer and the request were not linked to each other in any way, that is why I tried to use the start time. But you're right, if there were multiple requests at the same time for the same record, the problem would still be there. 

I now added `request_id` to `TransferMetadata` to link requests and transfers together, which should fix the issue for any new requests.

psaiz (approved):

